PR #2, opened 2018-09-30
Feature measurement webpage

SeboCode:


dist/benchmark/index.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>Measurement Webpage</title>
```
InverseIntegral: Rename to Benchmark

dist/benchmark/index.html
```diff
@@ -0,0 +1,29 @@
...
+    </head>
+    <body>
+        <section>
+            <header>
```
InverseIntegral: This header could be removed

dist/benchmark/index.html
```diff
@@ -0,0 +1,29 @@
...
+                <h1>Select an algorithm to measure it's speed</h1>
+            </header>
+            <select id="selected-algorithm">
+            </select>
```
InverseIntegral: Move closing tag to line 12

dist/benchmark/index.html
```diff
@@ -0,0 +1,29 @@
...
+            <input id="warmup-rounds" value="5" />
```
InverseIntegral: Add a description to the field

dist/benchmark/index.html
```diff
@@ -0,0 +1,29 @@
...
+            <input id="measure-rounds" value="100" />
```
InverseIntegral: Add a description to the field

src/benchmark/Measurement.js
```diff
@@ -0,0 +1,34 @@
...
+        // cross compile here
+        let wasmAlgorithm = jsAlgorithm; // change to cross compiled function
+        this.executeAlgorithm(wasmAlgorithm, this.warmupRounds);
+        let wasmTime = this.measureTime(wasmAlgorithm, this.measureRounds);
+
+        return {jsTime, wasmTime};
+    }
+
+    measureTime(algorithm) {
+        let startTime = Date.now();
```
InverseIntegral: Could startTime and endTime be const?

src/benchmark/Measurement.js
```diff
@@ -0,0 +1,34 @@
...
+    executeMeasurement(jsAlgorithm) {
+        this.executeAlgorithm(jsAlgorithm, this.warmupRounds);
+        let jsTime = this.measureTime(jsAlgorithm, this.measureRounds);
+
+        // cross compile here
+        let wasmAlgorithm = jsAlgorithm; // change to cross compiled function
+        this.executeAlgorithm(wasmAlgorithm, this.warmupRounds);
+        let wasmTime = this.measureTime(wasmAlgorithm, this.measureRounds);
+
+        return {jsTime, wasmTime};
```
InverseIntegral: Use a tuple instead of an object

src/benchmark/listeners.js
```diff
@@ -0,0 +1,44 @@
+import test_algorithm from './test_algorithm.js';
+import Measurement from './Measurement.js';
+
+const algorithms = {
+    test_algorithm: test_algorithm
+};
+
+function appendResult(result, log, selectedAlgorithm, warmupRounds, measureRounds) {
```
InverseIntegral: Only print the average time and variance of all rounds

src/benchmark/listeners.js
```diff
@@ -0,0 +1,44 @@
+import test_algorithm from './test_algorithm.js';
+import Measurement from './Measurement.js';
+
+const algorithms = {
```
InverseIntegral: A map from string to function would probably be better here

src/benchmark/Measurement.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    measureTime(algorithm) {
```
InverseIntegral: `measureTime` only returns the time of all executions.
It would be better to return an array with an element for each round.

dist/benchmark/index.html
InverseIntegral: Rename to benchmark

dist/benchmark/index.html
InverseIntegral: Add a description to the field

dist/benchmark/index.html
InverseIntegral: Add a description to the field

src/benchmark/Measurement.js
InverseIntegral: Use a tuple instead of an object

src/benchmark/Measurement.js
InverseIntegral: `measureTime` only returns the time of all executions.
It would be better to return an array with an element for each round.

dist/benchmark/index.html
InverseIntegral: This header could be removed

dist/benchmark/index.html
InverseIntegral: Move closing tag to line 12

src/benchmark/Measurement.js
InverseIntegral: Could `startTime` and `endTime` be const?

src/benchmark/listeners.js
InverseIntegral: Only print the average time and variance of all rounds

src/benchmark/listeners.js
InverseIntegral: A map from string to function would probably be better here

src/benchmark/listeners.ts
```diff
@@ -0,0 +1,97 @@
+import fibonacci from './cases/fibonacci';
```
InverseIntegral: You could rename this file to something like index.ts

src/benchmark/listeners.ts
```diff
@@ -0,0 +1,97 @@
...
+    log.innerText += 'Total JavaScript time: ' + totalJsTime + '\n';
+    log.innerText += 'Total WebAssembly time: ' + totalWasmTime + '\n';
+    log.innerText += 'Total time improvement: ' + (totalJsTime - totalWasmTime) + '\n';
+    log.innerText += 'Average JavaScript time: ' + mean(result[0]) + '\n';
+    log.innerText += 'Average WebAssembly time: ' + mean(result[1]) + '\n';
+    log.innerText += 'JavaScript median: ' + median(result[0]) + '\n';
+    log.innerText += 'WebAssembly median: ' + median(result[1]) + '\n';
+    log.innerText += 'JavaScript variance: ' + variance(result[0]) + '\n';
+    log.innerText += 'WebAssembly variance: ' + variance(result[1]) + '\n';
+    log.innerText += 'Warmup rounds amount: ' + warmupRounds + '\n';
```
InverseIntegral: Is this info necessary when running the benchmarks?

src/benchmark/listeners.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (values.length % 2 === 0) {
+        return values[half];
+    } else {
+        return (values[half - 1] + values[half]) / 2;
+    }
+}
+
+function variance(values: number[]): number {
+    const m = mean(values);
```
InverseIntegral: Rename the variable `m`

src/benchmark/listeners.ts
```diff
@@ -0,0 +1,97 @@
...
+
+function sum(value1: number, value2: number): number {
+    return value1 + value2;
+}
+
+function mean(values: number[]): number {
+    return values.reduce(sum, 0) / values.length;
+}
+
+function median(values: number[]): number {
```
InverseIntegral: Is the median calculation still necessary? Maybe the variance is sufficient.

src/benchmark/listeners.ts
SeboCode: The problem is, a variable can not be called the same name as a function. Therefore mean is not an option.

src/benchmark/listeners.ts
SeboCode: I would just leave it there. It doesn't hurt to have more data.

PR #5, opened 2018-10-02
Introduce visiting state and split visit function

InverseIntegral:


src/generator/VisitingState.ts
```diff
@@ -0,0 +1,11 @@
+import {Expression, Statement} from 'binaryen';
+
+class VisitingState {
```
SeboCode: Rename to VisitorState.

src/generator/VisitingState.ts
```diff
@@ -0,0 +1,11 @@
...
+
+    public expressions: Expression[] = [];
```
SeboCode: Rename to expressionStack.

PR #7, opened 2018-10-05
Implement branching

InverseIntegral:


src/generator/Generator.ts
```diff
@@ -111,10 +131,55 @@ class Generator {
...
+
+        if (immediateParent !== undefined && isIfStatement(immediateParent)) {
+            const parentBranches = state.branches.get(immediateParent);
+
+            if (parentBranches === undefined) {
+                throw Error('Parent branches are undefined');
+            }
+
+            parentBranches[1] = ifStatement;
+            state.branches.set(immediateParent, parentBranches);
```
SeboCode: Is this really necessary? You only change the value in an array. Therefore the reference should not have changed and the set call can savely be deleted.

src/generator/Generator.ts
```diff
@@ -111,10 +131,55 @@ class Generator {
...
+        } else {
+            state.statements.push(ifStatement);
+        }
     }
 
+    private visitBlockStatement(state: VisitorState, parent: TraversalAncestors) {
+        const block = this.module.block('', state.statements);
+        state.statements = []; // clear all current statements
+
+        const immediateParent = Generator.getImmediateParent(parent);
```
SeboCode: The upcoming code is more or less the exact same as the one in the visitIfStatement method. Extract them to one if possible.

src/generator/Generator.ts
```diff
@@ -15,13 +18,22 @@ import VisitorState from './VisitorState';
 
 class Generator {
 
-    private readonly module: Module;
+    private static getImmediateParent(parent: TraversalAncestors): Node | undefined {
```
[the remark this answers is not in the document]
InverseIntegral: Yes, static methods have to be at the top according to TSLint

src/generator/Generator.ts
InverseIntegral: I tried that using callbacks but it makes things a bit less readable. The common method would look like this:

```
private isParentIfStatement(parent: TraversalAncestors,
                            success: (immediateParent: IfStatement) => void,
                            failure: () => void) {
    const immediateParent = Generator.getImmediateParent(parent);

    if (immediateParent !== undefined && isIfStatement(immediateParent)) {
        success(immediateParent);
    } else {
        failure();
    }
}
```

And the call from `visitBlockStatement` would then use this new common method:

```
private visitBlockStatement(state: VisitorState, parent: TraversalAncestors) {
    const block = this.module.block('', state.statements);
    state.statements = []; // clear all current statements

    this.isParentIfStatement(parent, (immediateParent) => {
        const parentBranches = state.branches.get(immediateParent);

        if (parentBranches === undefined) {
            state.branches.set(immediateParent, [block, undefined]);
        } else {
            parentBranches[1] = block;
        }
    }, () => {
        state.body = block; // top statement block
    });
}
```
Is this more intention revealing? Or is there a better way to write the common part?

SeboCode: I thought more of something like this:
```
private updateImmediateParent(parent: TraversalAncestors,
    onParentBranchUndefined: (state: VisitorState, immediateParent: Node, block: Statement) => void,
    onNonIfStatement: (state: VisitorState, block: Statement) => void,
    state: VisitorState,
    block: Statement) {
    const immediateParent = Generator.getImmediateParent(parent);

    if (immediateParent !== undefined && isIfStatement(immediateParent)) {
        const parentBranches = state.branches.get(immediateParent);

        if (parentBranches === undefined) {
            onParentBranchUndefined(state, immediateParent, block);
        } else {
            parentBranches[1] = block;
        }
    } else {
        onNoIfStatement(state, block);
    }
}
```

The call from ```visitBlockStatement``` would then look like this:
```
private visitBlockStatement(state: VisitorState, parent: TraversalAncestors) {
    const block = this.module.block('', state.statements);
    state.statements = []; // clear all current statements
    updateImmediateParent(parent,
        (state, immediateParent, block) => { state.branches.set(immediateParent, [block, undefined]); },
        (state, block) => { state.body = block; },
        state,
        block);
}
```

The namings in my example aren't the best yet and I'm not yet sure if it is a good idea to implement it that way, but there wouldn't be any more duplicated code.

InverseIntegral: Unfortunately this doesn't work because `visitIfStatement` does more in the else branch than just `parentBranches[1] = block;` So the common method would require another callback. I think we should leave it as it is for the moment.

InverseIntegral: I see, this reduces about four lines of duplicated code. I'm not sure if it's worth the effort because the common function is quite complicated :confused:

SeboCode: Fair enough. We probably have to change this code later on anyways. Let's keep it this way for now.

PR #9, opened 2018-10-05
Implement mathematical and logical unary operators

SeboCode:


test/transpiler_test.ts
```diff
@@ -3,6 +3,27 @@ import Transpiler from '../src/Transpiler';
...
     describe('#transpile()', () => {
+        it('should handle unary plus', () => {
+            const {func} = Transpiler.transpile('function func(a) { return +a + +40 }');
+
+            expect(func(2)).to.equal(42);
+            expect(func(-2)).to.equal(38);
+        });
+
+        it('should handle unary minus', () => {
+            const {func} = Transpiler.transpile('function func(a) { return -a + -40 }');
```
InverseIntegral: Maybe a test with multiple unary minus signs would be helpful as well

PR #12, opened 2018-10-10
Implement variable declarations and assignments

InverseIntegral:


src/generator/generator.ts
```diff
@@ -1,7 +1,10 @@
-import {FunctionExpression, isIdentifier, LVal,} from '@babel/types';
+import {FunctionExpression} from '@babel/types';
 import {i32, Module} from 'binaryen';
+import DeclarationVisitor from './declaration_visitor';
 import GeneratorVisitor from './generator_visitor';
 
+type Mapping = Map<string, number>;
```
SeboCode: We should move this type definition to the visitor base class file, to prevent a coupling of the visitors to the generator.

src/generator/generator.ts
```diff
@@ -13,30 +16,26 @@ class Generator {
         const parameters = new Array(tree.params.length).fill(i32);
```
SeboCode: Move this down to line 24 and use parameterMapping.size to have a consistent solution.

src/generator/generator_visitor.ts
```diff
@@ -23,7 +24,7 @@ class GeneratorVisitor extends Visitor {
     constructor(module: Module, parameterMapping: Map<string, number>) {
```
SeboCode: Rename to variableMapping

src/visitor.ts
```diff
@@ -49,31 +73,57 @@ abstract class Visitor {
...
 
+    protected visitVariableDeclaration(node: VariableDeclaration) {
+        for (const declarator of node.declarations) {
+            this.visit(declarator);
+        }
     }
 
+    // noinspection TsLint
+    protected visitVariableDeclarator(node: VariableDeclarator) {
+
```
SeboCode: Add:
`if (node.init !== null) {
    this.visit(node.init);
}`
for consistency's sake

test/transpiler_test.ts
```diff
@@ -246,5 +246,41 @@ describe('Transpiler', () => {
...
+            expect(variables(100)).to.equal(10);
+        });
+
+        it('should handle a single variable with direct assignment', () => {
+            const {variables} = Transpiler.transpile('function variables() { var x = 10; return x; }');
+            expect(variables()).to.equal(10);
+        });
+
+        it('should handle multiple variables', () => {
+            const content = 'function variables(a) { var x, y; x = 10; y = 20; return x * y; }';
```
SeboCode: Remove superfluous argument

test/transpiler_test.ts
```diff
@@ -246,5 +246,41 @@ describe('Transpiler', () => {
...
+            const {variables} = Transpiler.transpile(content);
+            expect(variables(100)).to.equal(200);
+        });
+
+        it('should handle multiple variables with direct assignments', () => {
+            const content = 'function variables(a) { var x = 10, y = 20; return x * y; }';
```
SeboCode: Remove superfluous argument

test/transpiler_test.ts
InverseIntegral: I added the argument to check if the index for local variables starts at `1` instead of `0`

PR #16, opened 2018-10-15
Change statements to one dimensional array

InverseIntegral:
This removes the nested if statement workaround.

### DIFF
--- a/src/generator/generator_visitor.ts
+++ b/src/generator/generator_visitor.ts
@@ -24,7 +24,7 @@ class GeneratorVisitor extends Visitor {
     private readonly module: Module;
     private readonly variableMapping: Map<string, number>;
 
-    private statements: Statement[][] = [];
+    private statements: Statement[] = [];
     private expressions: Expression[] = [];
     private currentBlock: Statement;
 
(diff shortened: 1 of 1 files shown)


src/generator/generator_visitor.ts
```diff
@@ -299,7 +295,7 @@ class GeneratorVisitor extends Visitor {
     }
 
     private appendStatement(statement: Statement) {
```
SeboCode: Inline this method.

src/generator/generator_visitor.ts
```diff
@@ -195,13 +195,13 @@ class GeneratorVisitor extends Visitor {
         let elsePart;
 
         if (node.alternate !== null) {
+            const current = this.statements;
```
SeboCode: Rename to `parentStatements`.

src/generator/generator_visitor.ts
```diff
@@ -211,17 +211,13 @@ class GeneratorVisitor extends Visitor {
     }
 
     protected visitBlockStatement(node: BlockStatement) {
-        this.statements.push([]);
+        const current = this.statements;
```
SeboCode: Rename to `parentStatements`.

PR #19, opened 2018-10-15
Implement while loop support

SeboCode:


src/generator/generator_visitor.ts
```diff
@@ -220,6 +222,22 @@ class GeneratorVisitor extends Visitor {
         this.createSetLocal(node.left);
     }
 
+    protected visitWhileStatement(node: WhileStatement) {
+        super.visitWhileStatement(node);
+
+        const topBlockLabel = this.generateLabel();
+        const loopLabel = this.generateLabel();
+
+        const cancelBranch = this.module.br_if(topBlockLabel, this.negate(this.popExpression()));
```
InverseIntegral: The negation will probably have a negative performance impact. Maybe we can check if there's an alternate solution.

src/generator/generator_visitor.ts
```diff
@@ -220,6 +222,22 @@ class GeneratorVisitor extends Visitor {
...
+        const loopBranch = this.module.br(loopLabel);
+        const whilePart = this.popStatement();
+
+        const topBlock = this.module.block(topBlockLabel, [cancelBranch, whilePart, loopBranch]);
```
InverseIntegral: Maybe `topBlock` isn't the best name for this variable

src/generator/generator_visitor.ts
InverseIntegral: You could rename this variable to `conditionPart`

SeboCode: It is most likely that the negate has negative impact on the performance, but we should fix such things when we have proof for that assumptions like that.

PR #21, opened 2018-10-19
Restructure benchmark infrastructure

InverseIntegral:
- Uses a single method per benchmark case
- Allows for parameter passing to benchmarked function
- Logs benchmark results to console

### DIFF
--- a/src/benchmark/cases/newtons_method.d.ts
+++ b/src/benchmark/cases/newtons_method.d.ts
@@ -1,1 +1,1 @@
-export default function calculateNewtonsMethod(): void;
+export default function run(): void;
--- a/src/benchmark/cases/newtons_method.js
+++ b/src/benchmark/cases/newtons_method.js
@@ -1,24 +1,12 @@
-function F(x) {
-    return Math.pow(x, 2) - 612; // solve x^2 = 612
-}
-
-function f(x) {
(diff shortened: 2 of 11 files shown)


src/benchmark/cases/newtons_method.js
```diff
@@ -1,24 +1,12 @@
-function F(x) {
-    return Math.pow(x, 2) - 612; // solve x^2 = 612
-}
-
-function f(x) {
-    return 2 * x;
-}
-
-function solve(iterations, initial) {
+function newtons_method(iterations, initial) {
```
SeboCode: Rename to `newtonsMethod`

src/benchmark/cases/prime_counter.d.ts
```diff
@@ -1 +1 @@
-export default function executePrimeCounter(): void;
+export default function newtons_method(): void;
```
SeboCode: Rename to `isPrime`

src/benchmark/cases/newtons_method.d.ts
```diff
@@ -1 +1 @@
-export default function calculateNewtonsMethod(): void;
+export default function run(): void;
```
SeboCode: Rename to `newtonsMethod`

PR #23, opened 2018-10-23
Add new benchmark case

InverseIntegral:
- Added gcd benchmark
- Added while loops to increase measured time
- Added result check to benchmark execution

### DIFF
--- a/src/benchmark/cases/gcd.js
+++ b/src/benchmark/cases/gcd.js
@@ -1,0 +1,13 @@
+function gcd(a, b) {
+    while (b != 0) {
+        if (a > b) {
+            a = a - b;
+        } else {
+            b = b - a;
+        }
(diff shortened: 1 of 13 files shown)


src/benchmark/cases/gcd.js
```diff
@@ -0,0 +1,24 @@
+function gcd(a, b) {
+    while (b != 0) {
+        if (a > b) {
+            a = a - b;
```
SeboCode: Use shorthand assignment.

src/benchmark/cases/gcd.js
```diff
@@ -0,0 +1,24 @@
...
+        } else {
+            b = b - a;
```
SeboCode: Use shorthand assignment.

PR #25, opened 2018-10-26
Handle array import and array access

SeboCode:


src/transpiler.ts
```diff
@@ -13,7 +13,48 @@ class Transpiler {
...
+
+            return new WebAssembly.Instance(wasmModule, {
+                transpilerImports: {
+                    memory,
+                },
+            }).exports[functionName](...wasmParams);
+        };
+    }
+
+    private static fillMemory(params: any[], writeableMemory: Uint32Array) {
```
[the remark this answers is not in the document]
SeboCode: I thought about that as well. The problem is it would be inefficient, because I would need to loop over the params elements twice.

src/transpiler.ts
```diff
@@ -13,7 +13,48 @@ class Transpiler {
...
+        const wasmParams = params;
```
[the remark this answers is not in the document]
SeboCode: I made an error here. The `wasmParams` array should be a copy of the `params` array to not override any elements in it.

src/transpiler.ts
```diff
@@ -13,7 +13,51 @@ class Transpiler {
...
+                importObject = { transpilerImports: { memory } };
+            }
+
+            return new WebAssembly.Instance(wasmModule, importObject).exports[functionName](...wasmParams);
+        };
+    }
+
+    private static fillMemory(params: any[], writeableMemory: Uint32Array) {
+        // A copy of the params array is needed, to not override the content of the original one
+        const wasmParams = params.concat();
```
InverseIntegral: Add a test to check if the parameters were not modified

src/transpiler.ts
```diff
@@ -13,7 +13,51 @@ class Transpiler {
             throw new Error('The generated WebAssembly is invalid');
         }
 
-        return new WebAssembly.Instance(new WebAssembly.Module(module.emitBinary()), {}).exports;
+        const wasmModule = new WebAssembly.Module(module.emitBinary());
+
+        return (functionName: string, ...params: any[]) => {
+            let wasmParams = params;
```
InverseIntegral: This is probably not the best name for this variable. Maybe `rewrittenParameters` and `parameters` instead of `params` would be more intention revealing.

src/transpiler.ts
SeboCode: If no `concat()` is used, it would only modify the top level `params` array elements. The passed array does not get modified. Therefore it can not be tested from the outside. However, the reference to the passed array would be lost to us, making it impossible to later on write back the WebAssembly memory values.

PR #29, opened 2018-10-30
Combine array access logic and remove duplicate code

SeboCode:


src/generator/generator_visitor.ts
```diff
@@ -339,28 +328,40 @@ class GeneratorVisitor extends Visitor {
...
-
-            // @ts-ignore because store() returns an expression
-            this.statements.push(this.module.i32.store(0, 4, address, value));
+            this.setArrayElement(val, value);
         } else {
             throw new Error('Assignment to non-identifier or member expression');
         }
     }
 
+    private setLocal(identifier: Identifier, value: Expression) {
```
InverseIntegral: This method is only used once, inlining it would probably be better

src/generator/generator_visitor.ts
```diff
@@ -268,27 +269,15 @@ class GeneratorVisitor extends Visitor {
...
+            this.getArrayElement(node);
+        } else if (isIdentifier(node.property)) {
+            if (node.property.name === 'length') {
+                this.getArrayLength(node);
             } else {
-                throw new Error('property of static member expression was not an identifier');
+                throw new Error(`Unknown property ${node.property}`);
             }
+        } else {
+            throw new Error('The member access was neither an array access nor was it a property as an identifier');
```
InverseIntegral:
```suggestion
            throw new Error('The member access was neither a computer nor a static member expression');
```

By the way, this branch will never be taken because the two branches before already cover all cases. (See [here](https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md#memberexpression))

src/generator/generator_visitor.ts
```diff
@@ -370,9 +371,17 @@ class GeneratorVisitor extends Visitor {
 
         return index;
     }
+
+    private getAdress() {
```
InverseIntegral: Rename this method to `getArrayPointer()` because it's not the address that is returned

src/generator/generator_visitor.ts
```diff
@@ -339,28 +328,40 @@ class GeneratorVisitor extends Visitor {
...
+    }
+
+    private getArrayElement(node: MemberExpression) {
+        super.visitMemberExpression(node);
+
+        // The offset can not be used, because the memberaccess value can also be a mathematical term or a variable
+        this.expressions.push(this.module.i32.load(0, 4, this.getAdress()));
+    }
+
+    private getArrayLength(node: MemberExpression) {
```
InverseIntegral: Can't you use the `getAdress` method here as well?

SeboCode: Yes I could, if I push a constant `-1` onto the expression stack. The problem I have with such an implementation is that it would not improve readability. If the binaryen optimizer does not inline the multiplication of the two constant values we would even loose some performance.

src/generator/generator_visitor.ts
SeboCode: I remove it entirely then. There is no need for code that can never be reached.

PR #30, opened 2018-11-02
Refine benchmark measurements

InverseIntegral:


src/benchmark/measurement.ts
```diff
@@ -1,62 +1,92 @@
...
+    private static executeJS(algorithm: Function,
+                             args: any[],
+                             expectedResult: any,
+                             rounds: number): Timing[] {
+        const times: Timing[] = [];
 
         for (let i = 0; i < rounds; i++) {
-            const startTime: number = Date.now();
+            const start = performance.now();
             const result = algorithm(...args);
```
SeboCode: We need to work with a deep copy of the args array. Otherwise object parameters do not always have the initial values.

src/benchmark/measurement.ts
```diff
@@ -1,62 +1,92 @@
...
+    private static executeWasm(algorithm: Function,
+                             args: any[],
+                             expectedResult: any,
+                             rounds: number): Timing[] {
+        const times: Timing[] = [];
 
-        Measurement.executeAlgorithm(func[0], args, expectedResult, this.warmupRounds);
-        const jsTimes = Measurement.executeAlgorithm(func[0], args, expectedResult, this.measureRounds);
+        for (let i = 0; i < rounds; i++) {
+            const result = algorithm(...args);
```
SeboCode: We need to work with a deep copy of the args array. Otherwise object parameters do not always have the initial values.

src/transpiler.ts
```diff
@@ -1,9 +1,28 @@
...
+    performance = window.performance;
+}
+
+interface Timing {
+    compilationTime: number;
+    importTime: number;
+    executionTime: number;
+}
+
 class Transpiler {
```
SeboCode: I think it would be better to not use static methodes any longer to encapsulate the time values in an object. Otherwise if we have to expand this code it will get much more time consuming to avoid inconsistencies with the static members.

src/benchmark/measurement.ts
InverseIntegral: We don't support objects yet so this shouldn't be a problem at the moment

src/transpiler.ts
InverseIntegral: So you suggest changing both `transpile` and `getTiming` to non-static methods?

src/benchmark/measurement.ts
SeboCode: Well we support arrays and this problem can occur as soon as we have the export functionality implementet. We can postpone this until we need it. I just brought it up so it doesn't get forgotten.

src/transpiler.ts
SeboCode: Exactly.

InverseIntegral: The problem is that `transpile` does not use any state and should therefore be non-static.

src/benchmark/measurement.ts
InverseIntegral: Good, we will implement this when we work on the export functionality. Most of our benchmark cases will fill the arrays in the wasm code, so it shouldn't be a problem.

src/transpiler.ts
SeboCode: If we change it to a non-static version we could use a subclass for the time measurements. Therefore the production code doesn't get bloated with measurement code. The drawback is that one has to create a new `Transpiler` instance and not just call a static member function. But in my opinion this is the lesser evil

src/benchmark/index.ts
```diff
@@ -57,23 +57,23 @@ function variance(values: number[]): number {
     return mean(values.map((n) => Math.pow(n - meanValue, 2)));
 }
 
-function extractExecutionTime(timing: Timing): number {
+function extractExecutionTime(timing: Measurement): number {
```
SeboCode: Rename parameter to `measurement`.

src/benchmark/index.ts
```diff
@@ -57,23 +57,23 @@ function variance(values: number[]): number {
...
     return timing.executionTime;
 }
 
-function extractCompilationTime(timing: Timing): number {
+function extractCompilationTime(timing: Measurement): number {
```
SeboCode: Rename parameter to `measurement`.

src/benchmark/index.ts
```diff
@@ -57,23 +57,23 @@ function variance(values: number[]): number {
...
     return timing.compilationTime;
 }
 
-function extractImportTime(timing: Timing) {
+function extractImportTime(timing: Measurement) {
```
SeboCode: Rename parameter to `measurement`.

src/benchmark/index.ts
```diff
@@ -57,23 +57,23 @@ function variance(values: number[]): number {
...
     return timing.importTime;
 }
 
-function extractTotalTime(timing: Timing) {
+function extractTotalTime(timing: Measurement) {
```
SeboCode: Rename parameter to `measurement`.

PR #31, opened 2018-11-02
Improve quicksort benchmark

InverseIntegral:
Fills the array in the benchmark code by using a simple PRNG

### DIFF
--- a/src/benchmark/cases/quicksort.js
+++ b/src/benchmark/cases/quicksort.js
@@ -48,4 +48,54 @@ function quickSort(items, left, right) {
     return items;
 }
 
-module.exports = {quickSort, partition, swap};
+// LCG with GCCs parameters
+function fill(array) {
+    var i = 0;
+
+    var modulus = 0x80000000; // 2^31
+    var multiplier = 1103515245;
+    var increment = 12345;
+
+    var state = 42;
+
+    while (i < array.length) {
+        state = (multiplier * state + increment) % modulus;
+        array[i] = state;
+        i++;
+    }
+
+    return 0;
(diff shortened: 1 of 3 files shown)


src/benchmark/index.ts
```diff
@@ -35,12 +36,19 @@ const newtonsMethodFunc = {
     func: [newtonsMethodWhile, newtonsMethod],
 };
 
+const quicksortFunc = {
+    arguments: [new Array(1000000), 0, 99],
```
SeboCode: The second and third argument is not necessary, because the `quickSortWhile` function only uses the first one.

src/benchmark/cases/quicksort.js
```diff
@@ -48,4 +48,54 @@ function quickSort(items, left, right) {
...
+    var i = 0;
+
+    var modulus = 0x80000000; // 2^31
+    var multiplier = 1103515245;
+    var increment = 12345;
+
+    var state = 42;
+
+    while (i < array.length) {
+        state = (multiplier * state + increment) % modulus;
```
SeboCode: The multiplication creates an integer overflow, resulting in bad wasm times and wrong results.

InverseIntegral: Yes there are integer overflows because of the multiplication but the wasm time is not bad at all and the result is still correct. Maybe there is a PRNG that works with 32-bit integers

SeboCode: If the wasm time isn't bad because of the overflows and the result is correct then I don't see the need to change it.

PR #32, opened 2018-11-02
Implement sum array benchmark

SeboCode:


src/benchmark/cases/sum_array.js
```diff
@@ -0,0 +1,36 @@
...
+
+    while (i < array.length) {
+        sum += array[i];
+        i++;
+    }
+
+    return sum;
+}
+
+function fill(array) {
```
InverseIntegral: I think you should use the PRNG here

PR #34, opened 2018-11-03
Implement mergesort benchmark

SeboCode:


src/benchmark/cases/mergesort.js
```diff
@@ -0,0 +1,108 @@
...
+        workspaceArray[index] = array[indexRight];
+        indexRight++;
+        index++;
+    }
+
+    copyArray(array, workspaceArray, left, right);
+    return workspaceArray;
+}
+
+function isSorted(array) {
```
InverseIntegral: Is it possible to reuse the `isSorted` and the `fill` function of the other benchmark cases?

src/benchmark/cases/mergesort.js
```diff
@@ -0,0 +1,108 @@
...
+    mergeSort(array, workspaceArray, left, middle);
+    mergeSort(array, workspaceArray, middle, right);
+
+    merge(array, workspaceArray, left, middle, right);
+
+    return workspaceArray;
+}
+
+function merge(array, workspaceArray, left, middle, right) {
+    var indexLeft = left;
```
InverseIntegral: Rename this variable to `leftIndex`

src/benchmark/cases/mergesort.js
```diff
@@ -0,0 +1,108 @@
+function copyArray(destination, source, includingStartIndex, excludingEndIndex) {
```
InverseIntegral: Maybe rename the `includingStartIndex` and `excludingEndIndex` parameters to `fromIndex` and `toIndex`

src/benchmark/cases/mergesort.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    while (indexRight < right) {
+        workspaceArray[index] = array[indexRight];
+        indexRight++;
+        index++;
+    }
+
+    copyArray(array, workspaceArray, left, right);
+    return workspaceArray;
```
InverseIntegral: Is there a reason why you return the `workspaceArray` here?

SeboCode: Not really no. I just return something because it is necessary for the transpiler.

src/benchmark/cases/mergesort.js
SeboCode: No sadly not because JavaScript can not call functions that are defined in another file.

PR #35, opened 2018-11-04
Add time based tests

InverseIntegral:
@SeboCode can you check if those tests pass on your machine?


SeboCode:
Ran the mergesort test three times and it took about 15.5s each to pass on my machine. I increased the threshold for that test by one second.

PR #36, opened 2018-11-07
Handle array export and revise call wrapper

SeboCode:
All arrays are exported back to js. This is not yet done dynamically.

The call wrapper is now no longer just a function, but an object itself. It implements an easier version of the builder pattern.

The hook methods are now implemented in a separate object, which is passed as an argument to the transpiler.


src/benchmark/benchmark.ts
```diff
@@ -65,12 +68,10 @@ class Benchmark {
...
         const jsTimes = Benchmark.executeJS(func[0], args, expectedResult, measureRounds);
 
-        const wasmArgs = args.slice();
-        wasmArgs.unshift(func[0].name);
-
-        const wasmAlgorithm = this.transpiler.transpile(func.join(''));
-        this.executeWasm(wasmAlgorithm, wasmArgs, expectedResult, warmupRounds);
-        const wasmTimes = this.executeWasm(wasmAlgorithm, wasmArgs, expectedResult, measureRounds);
+        const callWrapper = transpiler.transpile(func.join(''));
+        const wasmAlgorithm = callWrapper.setCallFunctionName(func[0].name).call.bind(callWrapper);
```
InverseIntegral: Why is this `bind` call here necessary?

src/call_wrapper.ts
```diff
@@ -0,0 +1,106 @@
...
+            .map((array) => array.length + 1)
+            .reduce((accumulator, current) => accumulator + current, 0);
+
+        return Math.ceil((memoryElementCount * 4) / pageSize);
+    }
+
+    private readonly hook: TranspilerHook;
+    private readonly wasmModule: Module;
+
+    private callFunctionName: string;
```
InverseIntegral: `callFunctionName` is probably not the best name here. Maybe just `functionName`?

src/transpiler_hook.ts
```diff
@@ -0,0 +1,29 @@
+class TranspilerHook {
```
InverseIntegral: `TranspilerHook` is probably not the best name. If you want to name it like this then it would be `TranspilerHooks`

InverseIntegral: And using an interface here would probably be better. The only downside I see, is that we would need an empty implementation.

test/time_limited/node_benchmark_hook.ts
```diff
@@ -1,7 +1,7 @@
 import {performance} from 'perf_hooks';
-import Transpiler from '../../src/transpiler';
+import TranspilerHook from '../../src/transpiler_hook';
 
-class NodeBenchmarkTranspiler extends Transpiler {
+class NodeBenchmarkTranspiler extends TranspilerHook {
```
InverseIntegral: This class has to be renamed

test/time_limited/transpiler_test.ts
```diff
@@ -11,64 +11,68 @@ import {
...
     });
 
     describe('#transpile()', () => {
-        it('should run isPrime faster than 7 seconds', () => {
+        it('should setCallFunctionName isPrime faster than 7 seconds', () => {
             const exports = transpiler.transpile(isPrime.toString() + isPrimeWhile.toString());
 
-            expect(exports('isPrimeWhile', 46327)).to.equal(1);
-            expect(transpiler.getCompleteTime()).to.be.lessThan(7000);
+            expect(exports.setCallFunctionName('isPrimeWhile').call(46327)).to.equal(1);
```
InverseIntegral: Maybe rename exports in all tests to wrapper or something similar.

src/benchmark/benchmark.ts
SeboCode: Because we extract the call function and run it on it's own. The this context is therefore undefined. I changed it so that no longer the function but the callwrapper is passed to the execute method.

src/transpiler_hook.ts
SeboCode: If we use an interface, every hook class needs to implement all hook methods. I don't think this is a good idea.

README.md
```diff
@@ -9,10 +9,46 @@ clone the repository.
...
 
 ```javascript
-const exports = Transpiler.transpile('function add(a, b) { return a + b; }');
-const result = exports('add', 12, 30);
+const wrapper = new Transpiler().transpile('function add(a, b) { return a + b; }');
+wrapper.setFunctionName('add').call(21, 21);
 ```
+
+# API
+## Transpiler creation
```
InverseIntegral: Is it necessary to distinguish between the creation and the manipulation of the transpiler?

README.md
```diff
@@ -9,10 +9,46 @@ clone the repository.
...
+
+## Transpiler manipulation
+* Transpiler#**transpile**(content: `string`): `CallWrapper` <br />
+  Transpiles the specified content.
+
+## CallWrapper manipulation
+* CallWrapper#**setFunctionName**(functionName: `string`): `CallWrapper` <br />
+  Set the name of the initial function to be called.
+* CallWrapper#**call**(...parameters: `any[]`): `any` <br />
+  Call the initial function of the transpiled Wasm code.
```
InverseIntegral:
```suggestion
  Calls the initial function of the transpiled Wasm code.
```

README.md
```diff
@@ -9,10 +9,46 @@ clone the repository.
...
+
+## TranspilerHooks manipulation
+All methods are hook methods, which are called by the `Transpiler` and the `CallWrapper`. The default implementation of all these hook methods is empty, but it can be derived from and passed along to the `Transpiler`.
+
+* TranspilerHooks#**beforeCompilation**(): `void` <br />
+  Called before the transpilation process
```
InverseIntegral: Add full stops after those sentences.

README.md
```diff
@@ -9,10 +9,46 @@ clone the repository.
...
 ```javascript
-const exports = Transpiler.transpile('function add(a, b) { return a + b; }');
-const result = exports('add', 12, 30);
+const wrapper = new Transpiler().transpile('function add(a, b) { return a + b; }');
+wrapper.setFunctionName('add').call(21, 21);
 ```
+
+# API
+## Transpiler creation
+* new Transpiler(): `void` <br />
```
InverseIntegral: Why is only the return type marked as code?

README.md
```diff
@@ -9,10 +9,46 @@ clone the repository.
...
+  Transpiles the specified content.
+
+## CallWrapper manipulation
+* CallWrapper#**setFunctionName**(functionName: `string`): `CallWrapper` <br />
+  Set the name of the initial function to be called.
+* CallWrapper#**call**(...parameters: `any[]`): `any` <br />
+  Call the initial function of the transpiled Wasm code.
+
+## TranspilerHooks manipulation
+All methods are hook methods, which are called by the `Transpiler` and the `CallWrapper`. The default implementation of all these hook methods is empty, but it can be derived from and passed along to the `Transpiler`.
```
InverseIntegral:
```suggestion
The following methods are hook methods, which get called by the `Transpiler` and the `CallWrapper`. The default implementations of them are empty and can be overwritten.
```

src/transpiler_hook.ts
InverseIntegral: I don't think that this would be a big problem because in this case we want to describe the contract that the hook methods have to fulfil.

README.md
SeboCode: I had to change this a little bit, because `TranspilerHooks` now is an interface.

README.md
SeboCode: I don't think it looks that good if the entire line is marked as code.

src/transpiler_hooks.ts
```diff
@@ -1,4 +1,22 @@
...
+    beforeExport(): void;
+
+    afterExport(): void;
+
+    beforeImport(): void;
+
+    afterImport(): void;
+}
+
+class NullTranspilerHooks implements TranspilerHooks {
```
InverseIntegral: Move to separate file

InverseIntegral: And write the following methods like this:
```javascript
public beforeCompilation() {}
```

PR #37, opened 2018-11-10
Handle for loop

SeboCode:


src/visitor.ts
```diff
@@ -147,6 +155,22 @@ abstract class Visitor {
         this.visit(node.body);
     }
 
+    protected visitForStatement(node: ForStatement) {
```
InverseIntegral: I don't know if this should be the default behaviour. Wouldn't it be more consistent if you visit the `update` before visiting the `body` here?

src/generator/generator_visitor.ts
```diff
@@ -230,18 +231,24 @@ class GeneratorVisitor extends Visitor {
...
-        const endLabel = this.generateLabel();
-        const beginLabel = this.generateLabel();
+    protected visitForStatement(node: ForStatement) {
+        super.visitForStatement(node);
 
-        const conditionBranch = this.module.br_if(endLabel, this.module.i32.eqz(this.popExpression()));
-        const loopBranch = this.module.br(beginLabel);
-        const whilePart = this.popStatement();
+        let updateExpression;
+        let condition = this.module.i32.const(1);
```
InverseIntegral: Move this part after the following if statement

src/generator/generator_visitor.ts
```diff
@@ -230,18 +231,24 @@ class GeneratorVisitor extends Visitor {
...
 
-        const loopBlock = this.module.block(endLabel, [conditionBranch, whilePart, loopBranch]);
-        const whileStatement = this.module.loop(beginLabel, loopBlock);
+        if (node.update !== null) {
+            updateExpression = this.popStatement();
```
InverseIntegral: Shouldn't this be `this.popExpression()`?

test/transpiler/transpiler_for_test.ts
```diff
@@ -0,0 +1,149 @@
...
+    beforeEach(() => {
+        transpiler = new Transpiler();
+    });
+
+    describe('#transpile()', () => {
+        it('should handle simple for', () => {
+            const content = 'function loop(value, times) {' +
+                'for (var i = 0; i < times; i++) { value += 1; }' +
+                'return value; }';
+            const wrapper = transpiler.transpile(content);
```
InverseIntegral: Add the `setFunctionName` call on this line as well

test/transpiler/transpiler_for_test.ts
```diff
@@ -0,0 +1,149 @@
+import {expect} from 'chai';
+import Transpiler from '../../src/transpiler';
+
+describe('Transpiler', () => {
```
InverseIntegral: Add a test for for-loops without braces.

src/generator/generator_visitor.ts
SeboCode: No because this is either an `UpdateExpression`, an `AssignmentExpression` or a `ShorthandAssignment`. All of them are `Statements`

src/visitor.ts
SeboCode: It doesn't really matter, but it simplifies the generator visitor logic, because I need to be able to pop the `update` without poping the `body`. Otherwise I would need to store the `body` in a local variable and pass it to the `createLoop` helper method.

test/transpiler/transpiler_for_test.ts
SeboCode: There is already a for-loop test without braces.

PR #41, opened 2018-11-16
Handle array literals

SeboCode:


src/generator/array_literal_visitor.ts
```diff
@@ -0,0 +1,28 @@
+import {
+    ArrayExpression,
+    File,
+    isFunctionDeclaration,
+} from '@babel/types';
+import Visitor from '../visitor';
+
+class ArrayLiteralVisitor extends Visitor {
```
InverseIntegral: This visitor could also be part of the `DeclarationVisitor`

src/generator/declaration_visitor.ts
```diff
@@ -1,26 +1,57 @@
...
+        if (node.init !== null && isArrayExpression(node.init)) {
+            DeclarationVisitor.registerArray(node.id, node.init.elements.length, this.localArrayPointers);
+        }
+
         this.registerDeclaration(node.id, this.variables);
     }
 
-    private registerDeclaration(val: LVal, map: Mapping) {
+    protected visitAssignmentExpression(node: AssignmentExpression) {
+        if (isArrayExpression(node.right)) {
```
InverseIntegral: You could use visitArrayExpression directly here, can't you?
Because an array expression can also be part of a function call e.g. 
```javascript
otherFunction([1, 2, 3]);
```

src/generator/declaration_visitor.ts
```diff
@@ -1,26 +1,57 @@
...
+    VariableDeclarator,
+} from '@babel/types';
 import Visitor from '../visitor';
-type Mapping = Map<string, number>;
+
+type VariableMapping = Map<string, number>;
 
 class DeclarationVisitor extends Visitor {
 
+    private static memoryOffset: number = 0;
```
InverseIntegral: Make this variable non-static otherwise multiple transpile calls share the same `memoryOffset`

src/generator/generator_visitor.ts
```diff
@@ -30,16 +31,18 @@ class GeneratorVisitor extends Visitor {
...
     private readonly variableMapping: Map<string, number>;
+    private readonly localArrayPointers: Map<string, number>;
 
     private statements: Statement[] = [];
     private expressions: Expression[] = [];
 
     private labelCounter: number = 0;
 
-    constructor(module: Module, variableMapping: Map<string, number>) {
+    constructor(module: Module, variableMapping: Map<string, number>, localArrayPointers: Map<string, number>) {
```
InverseIntegral: Don't use `Map<string, number> `here. Instead reuse the type `VariableMapping`.

src/transpiler.ts
```diff
@@ -4,12 +4,14 @@ import NullTranspilerHooks from './null_transpiler_hooks';
...
 import TranspilerHooks from './transpiler_hooks';
 import Module = WebAssembly.Module;
+import {ArrayLiteralVisitor} from "./generator/array_literal_visitor";
 
 class Transpiler {
 
     private readonly hooks: TranspilerHooks;
 
     private wasmModule: Module;
+    private arrayLiteralMemorySize: number = 0;
```
InverseIntegral: Remove explicit type.

src/call_wrapper.ts
```diff
@@ -3,10 +3,10 @@ import Module = WebAssembly.Module;
 
 class CallWrapper {
 
-    private static fillMemory(parameters: any[], writeableMemory: Uint32Array) {
+    private static fillMemory(parameters: any[], writeableMemory: Uint32Array, initialSize: number) {
```
InverseIntegral: Make method non-static and use the `initialSize` field of the `CallWrapper` instead of passing it as a parameter.

src/call_wrapper.ts
```diff
@@ -44,25 +44,27 @@ class CallWrapper {
         }
     }
 
-    private static calculateInitialMemorySize(parameters: any[]): number {
+    private static calculateInitialMemorySize(parameters: any[], initialMemorySize: number): number {
```
InverseIntegral: Make method non-static and use the `initialSize` field of the `CallWrapper` instead of passing it as a parameter.

src/call_wrapper.ts
```diff
@@ -44,25 +44,27 @@ class CallWrapper {
...
 
     private readonly hooks: TranspilerHooks;
     private readonly wasmModule: Module;
+    private readonly arrayLiteralMemorySize: number;
 
     private functionName: string;
     private outParameters: any[];
 
-    public constructor(wasmModule: Module, hooks: TranspilerHooks) {
+    public constructor(wasmModule: Module, hooks: TranspilerHooks, arrayLiteralMemorySize: number) {
```
InverseIntegral: There are multiple inconsistent usages of this parameter. I found several names used for the same value:
- arrayLiteralMemorySize
- initialSize
- initialMemorySize

You should decide on one of them.

src/call_wrapper.ts
```diff
@@ -83,11 +85,15 @@ class CallWrapper {
 
         const hasArrayParameters = parameters.some((parameter) => parameter instanceof Array);
 
-        if (hasArrayParameters) {
-            const memory = new WebAssembly.Memory({ initial: CallWrapper.calculateInitialMemorySize(parameters) });
+        if (hasArrayParameters || this.arrayLiteralMemorySize !== 0) {
+            const memoryDescriptor = {
```
InverseIntegral: Change the name `memoryDescriptor` to `initialMemory`

src/generator/generator_visitor.ts
```diff
@@ -215,6 +218,11 @@ class GeneratorVisitor extends Visitor {
     protected visitVariableDeclarator(node: VariableDeclarator) {
         if (node.init !== null) {
             this.visit(node.init);
+
+            if (isArrayExpression(node.init)) {
```
InverseIntegral, line 223: This should be done in the `visitArrayExpression` method.

src/generator/generator_visitor.ts
```diff
@@ -226,6 +234,10 @@ class GeneratorVisitor extends Visitor {
             this.handleShorthandAssignment(node);
         }
 
+        if (isArrayExpression(node.right)) {
```
InverseIntegral, line 238: This should be done in the `visitArrayExpression` method.

src/generator/generator_visitor.ts
```diff
@@ -391,6 +403,35 @@ class GeneratorVisitor extends Visitor {
         return this.module.i32.store(0, 4, this.getPointer(), value);
     }
 
+    private createLocalArray(val: LVal, length: number) {
+        if (!isIdentifier(val)) {
+            throw new Error('Assigned to unknown variable');
```
InverseIntegral: This error message is not correct. The LValue was not an identifier.

src/generator/generator_visitor.ts
```diff
@@ -391,6 +403,35 @@ class GeneratorVisitor extends Visitor {
...
+        }
+
+        const memoryLocation = this.getArrayMemoryLocation(val.name);
+        const lengthAddress = this.module.i32.const(memoryLocation - 4);
```
InverseIntegral: Change the name `lengthAddress` to `lengthPointer`

src/generator/generator_visitor.ts
```diff
@@ -391,6 +403,35 @@ class GeneratorVisitor extends Visitor {
...
+        // @ts-ignore because store() returns an expression
+        this.statements.push(this.module.i32.store(0, 4, lengthAddress, this.module.i32.const(length)));
+
+        for (let i: number = length - 1; i >= 0; i--) {
```
InverseIntegral: Is it possible that the reverse order has a negative impact on the performance?

InverseIntegral: And you should remove the type of `i`.

src/generator/generator_visitor.ts
```diff
@@ -391,6 +403,35 @@ class GeneratorVisitor extends Visitor {
...
+            // @ts-ignore because store() returns an expression
+            this.statements.push(this.module.i32.store(i * 4, 4,
+                this.module.i32.const(memoryLocation), this.popExpression()));
+        }
+
+        this.expressions.push(this.module.i32.const(memoryLocation));
+    }
+
+    private getArrayMemoryLocation(name: string) {
```
InverseIntegral: Maybe change the function name to `getMemoryLocationOfArray`.

test/transpiler/transpiler_array_test.ts
```diff
@@ -196,5 +196,118 @@ describe('Transpiler', () => {
 
             expect(() => wrapper.setFunctionName('arrayExport').setOutParameters([71, 72]).call([71, 72])).to.throw();
         });
+
```
InverseIntegral, line 224: I couldn't find any test that uses an array expression in a function call:
```javascript
testFunction([1, 2, 3]);
```

Or one that uses an array expression directly:
```javascript
[1, 2, 3][1] + 2 == 4;
```

InverseIntegral (changes requested):

src/generator/declaration_visitor.ts
SeboCode: We decided to not allow this logic in our sublanguage.

test/transpiler/transpiler_array_test.ts
SeboCode: We decided to not allow this logic in our sublanguage.

src/generator/array_literal_visitor.ts
SeboCode: We decided to leave it split.

src/generator/declaration_visitor.ts
SeboCode: We decided to move the memory offset calculation into an own visitor to avoid the necessity for a static variable.

src/call_wrapper.ts
SeboCode: I named it `memoryDescriptor` becaus that is the name of the parameter name of `WebAssembly.Memory()`. I think it's a fitting name.

src/generator/generator_visitor.ts
SeboCode: That's not possible, because I need the variable name to create the local array at the correct memory position.

src/generator/generator_visitor.ts
SeboCode: That's not possible, because I need the variable name to create the local array at the correct memory position.

src/generator/generator_visitor.ts
SeboCode: I changed it to a normal for loop and now use the length to calculate the offset.

src/generator/generator_visitor.ts
```diff
@@ -391,6 +404,36 @@ class GeneratorVisitor extends Visitor {
         return this.module.i32.store(0, 4, this.getPointer(), value);
     }
 
+    private createLocalArray(val: LVal, length: number) {
+        if (!isIdentifier(val)) {
+            throw new Error('Array creation only allowed using an identifiers');
```
InverseIntegral:
```suggestion
            throw new Error('Array expression assigned to a non identifier');
```

InverseIntegral (changes requested):

src/generator/generator_visitor.ts
InverseIntegral: This doesn't solve the problem. The stores are still in descending order. I think we should leave it for now. But I would revert the change with the offset calculation.

SeboCode: The problem is that we will have at least one reverse for loop. Either in the `createLocalArray` or the `visitArrayExpression` method. If we push them in normal order, they need to be poped in reverse order. If we push them in reverse order, they can be poped in normal order.

src/call_wrapper.ts
```diff
@@ -122,6 +69,54 @@ class CallWrapper {
...
+                    writeableMemory[index++] = element;
+                }
+            }
+        }
+
+        return fixedParameters;
+    }
+
+    private readMemory(parameters: any[], fixedParameters: any[], readableMemory: Uint32Array) {
+
```
InverseIntegral: Remove superfluous empty line

src/generator/literal_memory_size_visitor.ts
```diff
@@ -0,0 +1,28 @@
+import {
+    ArrayExpression,
+    File,
+    isFunctionDeclaration,
+} from '@babel/types';
+import Visitor from '../visitor';
+
+class LiteralMemorySizeVisitor extends Visitor {
```
InverseIntegral: Rename to `ArrayExpressionVisitor`

src/generator/generator.ts
```diff
@@ -36,7 +39,11 @@ class Generator {
         return module;
     }
 
-    public static generateFunction(module: Module, tree: FunctionDeclaration, signatures: FunctionSignatures) {
+    public static generateFunction(module: Module,
+                                   tree: FunctionDeclaration,
+                                   arrayLiteralVisiotr: ArrayLiteralVisitor,
```
InverseIntegral: Typo in parameter name.

src/generator/generator.ts
```diff
@@ -36,7 +39,11 @@ class Generator {
...
+                                   signatures: FunctionSignatures) {
+
```
InverseIntegral, line 44: Maybe remove this empty line.

src/generator/array_literal_visitor.ts
```diff
@@ -1,26 +1,47 @@
...
+    VariableDeclarator,
 } from '@babel/types';
 import Visitor from '../visitor';
+import {VariableMapping} from './declaration_visitor';
 
 class ArrayLiteralVisitor extends Visitor {
 
-    private arrayLiteralMemorySize: number = 0;
+    private memoryOffset = 0;
+    private variables: VariableMapping;
```
InverseIntegral: This member is never used inside this class

InverseIntegral (changes requested):

InverseIntegral (approved):

PR #47, opened 2018-11-20
Handle boolean return type conversion

SeboCode:


src/call_wrapper.ts
```diff
@@ -107,6 +107,20 @@ class CallWrapper {
 
         this.hooks.afterExport();
 
+        return this.convertResult(result);
+    }
+
+    private convertResult(result: number) {
```
InverseIntegral: Maybe pass the currentSignature as a parameter.

src/call_wrapper.ts
```diff
@@ -107,6 +107,20 @@ class CallWrapper {
...
+        const signature = this.getCurrentSignature();
+
+        if (signature.returnType === WebAssemblyType.BOOLEAN) {
+            if (result !== 0 && result !== 1) {
+                throw new Error(`Return type of function ${this.functionName} was defined to be a boolean`);
```
InverseIntegral: State that the actual return type wasn't a boolean.

InverseIntegral: Also add a test that checks if this error gets thrown

InverseIntegral (changes requested):
In the `transpiler_array_test` file the test "should handle out parameter with no call parameters" should use boolean as return type.

In the `transpiler_variable_test` file there is no variable with a boolean assigned.

In the `transpiler_if_statement_test` file there are several tests that could use boolean as return type.

SeboCode:
I don't think it we should change the return value of the test "should handle out parameter with no call parameters" in the `transpiler_array_test` file to a boolean value. It's not necessary to always use a boolean whenever possible.

test/transpiler/transpiler_function_test.ts
```diff
@@ -81,5 +81,19 @@ describe('Transpiler', () => {
...
+
+        it('should handle function call with not matching return type conversion', () => {
+            const content = 'function func(value) { return value; }';
+
+            const wrapper = transpiler
+                .setSignature('func', WebAssemblyType.BOOLEAN, WebAssemblyType.INT_32)
+                .transpile(content);
+            wrapper.setFunctionName('func');
+
+            expect(wrapper.call(1)).to.equal(true);
```
InverseIntegral, line 93: Strictly speaking this and the following expect call should fail. The type inference could determine that the returned value is an int32 instead of a boolean.

InverseIntegral (approved):

PR #48, opened 2018-11-23
Add type inference

InverseIntegral:


src/generator/type_inference_visitor.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    private getTypeOfIdentifier(identifier: Identifier) {
+        for (const [key, value] of this.expressionTypes) {
+            if (isIdentifier(key) && key.name === identifier.name) {
+                return value;
+            }
+        }
+    }
+
+    private initializeTypes(tree: FunctionDeclaration, signature: FunctionSignature) {
```
SeboCode: Rename this method to `initializeParameterTypes`.

src/generator/type_inference_visitor.ts
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            this.expressionTypes.set(node.id, rightSideType);
+        }
+    }
+
+    protected visitAssignmentExpression(node: AssignmentExpression): void {
+        super.visit(node.right);
+
+        if (isIdentifier(node.left)) {
```
SeboCode, line 148: This is the same logic as the one in `visitVariableDeclarator`. Therefore it should be combined.

SeboCode (changes requested):

src/generator/type_inference_visitor.ts
InverseIntegral: The combined method would be too generic. It would require at least three parameters which isn't better than what we have now.

SeboCode (approved):

PR #49, opened 2018-11-24
Handle double as literals and in variables

SeboCode:


src/generator/wasm_type.ts
```diff
@@ -14,19 +15,37 @@ function toBinaryenType(type: WebAssemblyType) {
...
+        case WebAssemblyType.FLOAT_64:
+            return f64;
         case WebAssemblyType.BOOLEAN:
             return i32;
         default:
             throw new Error(`Unknown type ${WebAssemblyType[type]}`);
     }
 }
 
+function getWebAssemblyType(value: any) {
```
InverseIntegral: This function is only used once. Maybe there is a better solution for this.

src/generator/type_inference_visitor.ts
```diff
@@ -66,7 +72,7 @@ class TypeInferenceVisitor extends Visitor {
     protected visitNumericLiteral(node: NumericLiteral) {
         super.visitNumericLiteral(node);
 
-        this.expressionTypes.set(node, WebAssemblyType.INT_32);
+        this.expressionTypes.set(node, getWebAssemblyType(node.value));
```
InverseIntegral: The function `getWebassemblyType` also checks for arrays and booleans which is not necessary here.

src/generator/type_inference_visitor.ts
```diff
@@ -1,32 +1,38 @@
...
 } from '@babel/types';
 import Visitor from '../visitor';
 import {FunctionSignature, FunctionSignatures} from './generator';
-import {WebAssemblyType} from './wasm_type';
+import {getWebAssemblyType, WebAssemblyType} from './wasm_type';
 
 class TypeInferenceVisitor extends Visitor {
 
     private signatures: FunctionSignatures;
     private expressionTypes = new Map<Expression, WebAssemblyType>();
```
InverseIntegral: Use a type declaration for this map type.

src/generator/type_inference_visitor.ts
```diff
@@ -1,32 +1,38 @@
...
+    private variableTypes: WebAssemblyType[] = [];
```
InverseIntegral: Maybe use a type declaration for this array type.

src/generator/type_inference_visitor.ts
```diff
@@ -152,6 +160,18 @@ class TypeInferenceVisitor extends Visitor {
...
+            this.expressionTypes.set(node.left, rightSideType);
+            this.variableTypes.push(rightSideType);
+        } else if (isMemberExpression(node.left)) {
+            super.visit(node.left);
+
+            const rightSideType = this.expressionTypes.get(node.right);
+
+            if (rightSideType === undefined) {
+                throw new Error(`Unknown type for right side of
+                    assignment to ${(node.left.object as Identifier).name}`);
```
InverseIntegral: `node.left.object` could also be a non-identifier and this would result in `undefined`

src/generator/type_inference_visitor.ts
InverseIntegral: The order in which the variable types are pushed to this array is important. Are you sure that it's always correct?

src/generator/generator_visitor.ts
```diff
@@ -413,6 +420,25 @@ class GeneratorVisitor extends Visitor {
...
+        } else {
+            return this.module.i32;
+        }
+    }
+
+    private getExpressionType(expression: BabelExpression) {
+        const type = this.expressionTypes.get(expression);
+
+        if (type === undefined) {
+            throw new Error('Expression type not defined');
```
InverseIntegral: Add the type of the BabelExpression to the error message.

src/generator/generator_visitor.ts
```diff
@@ -413,6 +420,25 @@ class GeneratorVisitor extends Visitor {
         return 'label_' + this.labelCounter++;
     }
 
+    private getOperationType(expression: BabelExpression) {
+        const type = this.getExpressionType(expression);
+
+        if (type === WebAssemblyType.FLOAT_64) {
```
InverseIntegral: Move this part to the `WebAssemblyType` enum declaration.

src/generator/generator_visitor.ts
```diff
@@ -51,11 +58,11 @@ class GeneratorVisitor extends Visitor {
     protected visitIdentifier(node: Identifier) {
         const index = this.getVariableIndex(node.name);
 
-        this.expressions.push(this.module.getLocal(index, i32));
+        this.expressions.push(this.module.getLocal(index, toBinaryenType(this.getExpressionType(node))));
     }
 
     protected visitNumericLiteral(node: NumericLiteral) {
-        this.expressions.push(this.module.i32.const(node.value));
+        this.expressions.push(this.getOperationType(node).const(node.value));
```
InverseIntegral: Rewrite to:

```javascript
const type = this.getExpressionType(expression);
this.expressions.push(getOperationType(type).const(node.value));
```

InverseIntegral (changes requested):

src/generator/generator_visitor.ts
SeboCode, line 427: This doesn't work, because it accesses a property of the class.

src/generator/type_inference_visitor.ts
```diff
@@ -197,6 +197,17 @@ class TypeInferenceVisitor extends Visitor {
             this.expressionTypes.set(parameter, signature.parameterTypes[index]);
         });
     }
+
+    private updateVariableTypes(identifier: Identifier, rightSideType: WebAssemblyType) {
+        const currentValue = this.variableTypes.get(identifier.name);
```
InverseIntegral: Introduce a variable for `indentifier.name`

src/generator/type_inference_visitor.ts
```diff
@@ -197,6 +197,17 @@ class TypeInferenceVisitor extends Visitor {
...
+
+        if (currentValue !== undefined && currentValue !== rightSideType) {
```
InverseIntegral: Use a `contains` call instead of the `undefined` comparison.

src/generator/generator.ts
```diff
@@ -83,6 +83,22 @@ class Generator {
...
+        const types = [];
+
+        for (const key of variableMapping.keys()) {
+            const webAssemblyType = variableTypes.get(key);
+
+            if (webAssemblyType === undefined) {
+                throw new Error(`Type could not be infered for variable ${key}`);
+            }
+
+            types.push(toBinaryenType(webAssemblyType));
```
InverseIntegral: Maybe use 

```javascript
types.splice(variableMapping.get(key), 0, toBinaryenType(webAssemblyType);
```

here to ensure the correct order of the `types` array.

InverseIntegral (changes requested):

src/generator/generator.ts
SeboCode: I found another solution that imo looks better.

src/generator/type_inference_visitor.ts
```diff
@@ -200,11 +200,14 @@ class TypeInferenceVisitor extends Visitor {
...
+        if (this.variableTypes.has(variableName)) {
+            const currentValue = this.variableTypes.get(variableName);
+
+            if (currentValue !== rightSideType) {
+                throw new Error(`Tried to change the value type of ${variableName}
                 from ${currentValue} to ${rightSideType}`);
+            }
         }
 
         this.variableTypes.set(variableName, rightSideType);
```
InverseIntegral: Shouldn't this be inside an else branch?

src/generator/type_inference_visitor.ts
```diff
@@ -200,11 +200,14 @@ class TypeInferenceVisitor extends Visitor {
 
     private updateVariableTypes(identifier: Identifier, rightSideType: WebAssemblyType) {
```
InverseIntegral: Maybe rename the method to `updateVariableType`

src/generator/generator.ts
```diff
@@ -84,19 +84,10 @@ class Generator {
...
-            }
-
-            types.push(toBinaryenType(webAssemblyType));
-        }
-
-        return types;
+        return [...variableMapping.entries()]
+            .sort((first, second) => first[1] - second[1])
+            .map((entry) => variableTypes.get(entry[0]))
+            .map(toBinaryenType);
```
InverseIntegral, line 90: This could thow an error because of `undefined` entries. Is is possible to catch them before calling `toBinaryenType`?

InverseIntegral (changes requested):

src/generator/type_inference_visitor.ts
SeboCode: I found a bug with this current implementation that I have to fix. So this is rewritten either way.

src/generator/generator.ts
SeboCode: This should never happen either way, but yes I can add a filter for it.

InverseIntegral (approved):

PR #51, opened 2018-11-25
Handle double in expressions

SeboCode:


src/generator/wasm_type.ts
```diff
@@ -34,6 +34,14 @@ function getNumberType(value: number) {
     }
 }
 
+function getCommonNumberType(leftType: WebAssemblyType, rightType: WebAssemblyType) {
```
InverseIntegral: Maybe rename `leftType` and `rightType` to `first` and `second`.

InverseIntegral: Is this always the case for our operations? That a double on either side turns the type of the whole operation into a double.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,86 @@ class GeneratorVisitor extends Visitor {
...
-        const left = this.popExpression();
+        let right = this.popExpression();
+        let left = this.popExpression();
+
+        const rightType = this.getExpressionType(node.right);
+        const leftType = this.getExpressionType(node.left);
+        const commonNumberType = getCommonNumberType(leftType, rightType);
+
+        if (leftType !== commonNumberType) {
+            left = this.module.f64.convert_s.i32(left);
```
InverseIntegral: This depends on the implementation of `getCommonNumberType` and is not really dynamic. What if `getCommonNumberType` supports more number types later on?

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,86 @@ class GeneratorVisitor extends Visitor {
...
+                this.expressions.push(operationsInstance.sub(left, right));
                 break;
             case '*':
-                this.expressions.push(this.module.i32.mul(left, right));
+                this.expressions.push(operationsInstance.mul(left, right));
                 break;
             case '/':
-                this.expressions.push(this.module.i32.div_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.div_s(left, right));
```
InverseIntegral: Wouldn't `2 / 3` result in `1` which isn't the same as in JS? Or is this documented in the language set?

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,86 @@ class GeneratorVisitor extends Visitor {
...
+                    this.expressions.push(this.module.f64.div(left, right));
+                }
+
                 break;
             case '%':
-                this.expressions.push(this.module.i32.rem_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.rem_s(left, right));
+                } else {
+                    throw new Error('Modulo is not allowed with float values');
```
InverseIntegral, line 130: Why isn't this allowed? JavaScript allows it on its numbers.

src/generator/generator_visitor.ts
```diff
@@ -321,21 +371,44 @@ class GeneratorVisitor extends Visitor {
 
     private handleShorthandAssignment(node: AssignmentExpression) {
         this.visit(node.left);
-        const currentValue = this.popExpression();
-        const assignedValue = this.popExpression();
+        let currentValue = this.popExpression();
+        let assignedValue = this.popExpression();
+
+        if (!isIdentifier(node.left) && !isMemberExpression(node.left)) {
+            throw new Error('Shorthand assignment only allowed on an identifier or an array member');
```
InverseIntegral, line 341: Are there tests for this new potential error?

src/generator/type_inference_visitor.ts
```diff
@@ -186,6 +177,18 @@ class TypeInferenceVisitor extends Visitor {
                 return value;
             }
         }
+
+        throw new Error(`Unknown type for identifier ${identifier.name}`);
```
InverseIntegral, line 201: Is there a test that covers this error?

src/generator/type_inference_visitor.ts
```diff
@@ -157,26 +155,19 @@ class TypeInferenceVisitor extends Visitor {
         super.visit(node.right);
 
         if (isIdentifier(node.left)) {
-            const rightSideType = this.expressionTypes.get(node.right);
+            let rightSideType = this.getTypeOfExpression(node.right);
 
-            if (rightSideType === undefined) {
-                throw new Error(`Unknown type for right side of assignment to ${node.left.name}`);
+            if (node.operator !== '=') {
+                rightSideType = getCommonNumberType(this.getTypeOfIdentifier(node.left), rightSideType);
```
InverseIntegral: Wouldn't this change the type of the right side?

InverseIntegral (changes requested):
I haven't reviewed the new tests yet.

src/generator/wasm_type.ts
SeboCode: Yes this holds for all binary and unary operators and therefore also for all shorthand assignment operators.

src/generator/generator_visitor.ts
SeboCode: Yes this is a different behavior than in JS. But I think that because of the reason that wasm is strongly typed, this behavior is and should be expected. It is, after all, a different language.

Good point, I don't think this is documented yet. I will add it when I'm describing the double type.

src/generator/generator_visitor.ts
SeboCode: I don't know why, but the wasm specification only allows modulo on integer values.

src/generator/generator_visitor.ts
SeboCode: No it is never tested, because our language specification doesn't allow any other expressions on the left side of a shorthand assignment. Therefore it is almost impossible to test and to make sure, that the error was thrown at this location.

This check is only there, because `node.left` is used as a parameter of the `getExpressionType` call. Is there a better solution for this?

src/generator/type_inference_visitor.ts
SeboCode: No. I could test it, but the test would not realy represent the reason why this check exists. I added the test anyway though.

src/generator/type_inference_visitor.ts
SeboCode: No the type of the right side is not changed during an assignment.

src/generator/generator_visitor.ts
InverseIntegral: Okay, then we have to document this in our language set and add a test case for this error.

SeboCode: There is already a test for that. It's called `should handle double modulo`

src/generator/generator_visitor.ts
SeboCode: I tried to test it, but the problem is, that identifier and member expression are all LValues of our language set that are supported. Any other LValue would already throw an error in the `visit` method, because it can not be handled.

src/generator/type_inference_visitor.ts
SeboCode: The discussed optimization of setting the type of the right node doesn't work, because array member access expressions are allowed to be on the right side of a shorthand assignment. Their types must be correct, in order to call the `load` method on the correct operations instance in the generator visitor.

InverseIntegral: `getArrayElement` of the `GeneratorVisitor` doesn't distinguish between int and double at the moment so this shouldn't be a problem.

src/generator/wasm_type.ts
```diff
@@ -34,6 +34,14 @@ function getNumberType(value: number) {
     }
 }
 
+function getCommonNumberType(first: WebAssemblyType, second: WebAssemblyType) {
```
InverseIntegral, line 37: Maybe write a test for this function

src/generator/type_inference_visitor.ts
```diff
@@ -45,7 +47,9 @@ class TypeInferenceVisitor extends Visitor {
         let type;
 
         if (['+', '-', '/', '%', '*'].includes(operator)) {
-            type = WebAssemblyType.INT_32;
+            const leftType = this.getTypeOfExpression(node.left);
+            const rightType = this.getTypeOfExpression(node.right);
```
InverseIntegral, line 51: Add a newline after this one.

src/generator/type_inference_visitor.ts
```diff
@@ -186,6 +180,18 @@ class TypeInferenceVisitor extends Visitor {
...
         }
+
+        throw new Error(`Unknown type for identifier ${identifier.name}`);
+    }
+
+    private getTypeOfExpression(expression: Expression) {
+        const type = this.expressionTypes.get(expression);
+
+        if (type === undefined) {
+            throw new Error(`The type of expression ${expression.type} could not be infered`);
```
InverseIntegral, line 208: Is there a test case for this error?

src/generator/type_inference_visitor.ts
```diff
@@ -157,26 +156,21 @@ class TypeInferenceVisitor extends Visitor {
         super.visit(node.right);
 
         if (isIdentifier(node.left)) {
```
InverseIntegral, line 159: Maybe split those two branches into different methods.

src/generator/type_inference_visitor.ts
```diff
@@ -198,23 +204,31 @@ class TypeInferenceVisitor extends Visitor {
...
+                    throw new Error(`Tried to change the value type of ${name}
+                    from ${WebAssemblyType[currentValue]} to ${WebAssemblyType[type]}`);
                 }
             }
-        } else {
-            this.variableTypes.set(variableName, rightSideType);
+        }
+    }
+
+    private updateVariableType(identifier: Identifier, type: WebAssemblyType) {
```
InverseIntegral: Maybe change the name of the method to `updateIdentifierType` and rename `variableTypes` to `identifierTypes`.

src/generator/generator_visitor.ts
```diff
@@ -87,7 +88,9 @@ class GeneratorVisitor extends Visitor {
                 this.expressions.push(operand);
                 break;
             case '-':
-                this.expressions.push(this.module.i32.sub(this.module.i32.const(0), operand));
+                const type = this.getExpressionType(node);
+                const operationsInstance = this.getOperationsInstance(type);
```
InverseIntegral: Add newline after this line.

InverseIntegral: And rename `operationsInstance` to `instance`.

src/generator/generator_visitor.ts
```diff
@@ -440,6 +501,20 @@ class GeneratorVisitor extends Visitor {
 
         return type;
     }
+
+    private convertType(expression: Expression, from: WebAssemblyType, to: WebAssemblyType) {
+        if (from !== to) {
+            const operationsInstance = this.getOperationsInstance(to) as F64Operations;
```
InverseIntegral: This assumes that `to` is always `WebAssemblyType.FLOAT_64`

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
+        let left = this.popExpression();
+
+        const rightType = this.getExpressionType(node.right);
+        const leftType = this.getExpressionType(node.left);
+        const commonNumberType = getCommonNumberType(leftType, rightType);
+
+        left = this.convertType(left, leftType, commonNumberType);
+        right = this.convertType(right, rightType, commonNumberType);
+
+        const operationsInstance = this.getOperationsInstance(commonNumberType);
```
InverseIntegral: Rename to `instance`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
                 break;
             case '*':
-                this.expressions.push(this.module.i32.mul(left, right));
+                this.expressions.push(operationsInstance.mul(left, right));
                 break;
             case '/':
-                this.expressions.push(this.module.i32.div_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.div_s(left, right));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
+                    this.expressions.push(this.module.f64.div(left, right));
+                }
+
                 break;
             case '%':
-                this.expressions.push(this.module.i32.rem_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.rem_s(left, right));
+                } else {
```
InverseIntegral, line 129: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
                 break;
             case '!=':
-                this.expressions.push(this.module.i32.ne(left, right));
+                this.expressions.push(operationsInstance.ne(left, right));
                 break;
             case '<':
-                this.expressions.push(this.module.i32.lt_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.lt_s(left, right));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
+                    this.expressions.push(this.module.f64.lt(left, right));
+                }
+
                 break;
             case '<=':
-                this.expressions.push(this.module.i32.le_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.le_s(left, right));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
+                    this.expressions.push(this.module.f64.le(left, right));
+                }
+
                 break;
             case '>':
-                this.expressions.push(this.module.i32.gt_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.gt_s(left, right));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -100,42 +103,81 @@ class GeneratorVisitor extends Visitor {
...
+                    this.expressions.push(this.module.f64.gt(left, right));
+                }
+
                 break;
             case '>=':
-                this.expressions.push(this.module.i32.ge_s(left, right));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.ge_s(left, right));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

src/generator/generator_visitor.ts
```diff
@@ -166,12 +208,15 @@ class GeneratorVisitor extends Visitor {
         const currentValue = this.popExpression();
         let updatedValue;
 
+        const type = this.getExpressionType(node);
+        const operationsInstance = this.getOperationsInstance(type);
```
InverseIntegral: Rename to `instance` and inline the `type` variable.

src/generator/generator_visitor.ts
```diff
@@ -321,21 +366,39 @@ class GeneratorVisitor extends Visitor {
...
+        }
+
+        const rightType = this.getExpressionType(node.right);
+        const leftType = this.getExpressionType(node.left);
+        const commonNumberType = getCommonNumberType(leftType, rightType);
+
+        currentValue = this.convertType(currentValue, leftType, commonNumberType);
+        assignedValue = this.convertType(assignedValue, rightType, commonNumberType);
+
+        const operationsInstance = this.getOperationsInstance(commonNumberType);
```
InverseIntegral: Rename to `instance`.

src/generator/generator_visitor.ts
```diff
@@ -321,21 +366,39 @@ class GeneratorVisitor extends Visitor {
...
                 break;
             case '*=':
-                this.expressions.push(this.module.i32.mul(currentValue, assignedValue));
+                this.expressions.push(operationsInstance.mul(currentValue, assignedValue));
                 break;
             case '/=':
-                this.expressions.push(this.module.i32.div_s(currentValue, assignedValue));
+                if (commonNumberType === WebAssemblyType.INT_32) {
+                    this.expressions.push(this.module.i32.div_s(currentValue, assignedValue));
+                } else {
```
InverseIntegral: Replace with an else if and throw an error in case it's neither `INT_32` nor `FLOAT_64`.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
                 .setSignature('double', WebAssemblyType.FLOAT_64)
                 .transpile('function double() { return 1.34; } ');
 
-            expect(wrapper.setFunctionName('double').call()).to.equal(1.34);
+            expect(wrapper.setFunctionName('double').call()).to.closeTo(1.34, 0.001);
```
InverseIntegral: `to.equal()` also works here. Is there a reason why you used `closeTo()`?

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
         });
 
         it('should handle double parameter', () => {
             const wrapper = transpiler
                 .setSignature('double', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
                 .transpile('function double(value) { return value; } ');
 
-            expect(wrapper.setFunctionName('double').call(3.14)).to.equal(3.14);
+            expect(wrapper.setFunctionName('double').call(3.14)).to.closeTo(3.14, 0.001);
```
InverseIntegral: Add more calls of the `double` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double unary plus', () => {
+            const wrapper = transpiler
+                .setSignature('double', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function double(a) { return +a; }');
+            wrapper.setFunctionName('double');
+
+            expect(wrapper.call(31.412)).to.closeTo(31.412, 0.001);
+            expect(wrapper.call(-231.4129)).to.closeTo(-231.4129, 0.00001);
```
InverseIntegral: Add a call with argument `0.0`.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double unary minus', () => {
+            const wrapper = transpiler
+                .setSignature('double', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function double(a) { return -a ; }');
+            wrapper.setFunctionName('double');
+
+            expect(wrapper.call(132.319)).to.closeTo(-132.319, 0.0001);
+            expect(wrapper.call(-18.95)).to.closeTo(18.95, 0.001);
```
InverseIntegral: Add a call with argument `0.0` and `-0.0`.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+            expect(wrapper.call(-9.49)).to.closeTo(9.49, 0.001);
+        });
+
+        it('should handle double pre increment', () => {
+            const content = 'function preInc(a) { ++a; return a; }';
+            const wrapper = transpiler
+                .setSignature('preInc', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile(content);
+
+            expect(wrapper.setFunctionName('preInc').call(12.34)).to.closeTo(13.34, 0.001);
```
InverseIntegral: Add more calls of the `preInc` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double post increment', () => {
+            const content = 'function postInc(a) { a++; return a; }';
+            const wrapper = transpiler
+                .setSignature('postInc', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile(content);
+
+            expect(wrapper.setFunctionName('postInc').call(4.92)).to.closeTo(5.92, 0.001);
```
InverseIntegral: Add more calls of the `postInc` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double pre decrement', () => {
+            const content = 'function preDec(a) { --a; return a; }';
+            const wrapper = transpiler
+                .setSignature('preDec', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile(content);
+
+            expect(wrapper.setFunctionName('preDec').call(93.89)).to.closeTo(92.89, 0.001);
```
InverseIntegral: Add more calls of the `preDec` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double post decrement', () => {
+            const content = 'function postDec(a) { a--; return a; }';
+            const wrapper = transpiler
+                .setSignature('postDec', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile(content);
+
+            expect(wrapper.setFunctionName('postDec').call(28.19)).to.closeTo(27.19, 0.001);
```
InverseIntegral: Add more calls of the `postDec` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+                .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32)
+                .transpile(content);
+            wrapper2.setFunctionName('add');
+
+            expect(wrapper2.call(5.31, 4)).to.closeTo(9.31, 0.001);
+            expect(wrapper2.call(14.1, -3)).to.closeTo(11.1, 0.001);
+            expect(wrapper2.call(-23.4, 35)).to.closeTo(11.6, 0.001);
+            expect(wrapper2.call(-9.5, -2)).to.closeTo(-11.5, 0.001);
+        });
+
```
InverseIntegral, line 183: Add tests where you add a parameter to a double literal.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        it('should handle double addition with array', () => {
+            const wrapper = transpiler
+                .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY)
+                .transpile('function add(a, b) { return a + b[0]; }');
+            wrapper.setFunctionName('add');
+
+            expect(wrapper.call(1.13, [1, 2])).to.closeTo(2.13, 0.001);
```
InverseIntegral: Add more calls to the `add` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+
+            const wrapper2 = transpiler
+                .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64)
+                .transpile('function add(a, b) { return a[0] + b; }');
+            wrapper2.setFunctionName('add');
+
+            expect(wrapper2.call([1, 2], 2.4)).to.closeTo(3.4, 0.001);
```
InverseIntegral: Add more calls to the `add` function.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+                .setSignature('sub', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32)
+                .transpile(content);
+            wrapper2.setFunctionName('sub');
+
+            expect(wrapper2.call(5.31, 4)).to.closeTo(1.31, 0.001);
+            expect(wrapper2.call(14.1, -3)).to.closeTo(17.1, 0.001);
+            expect(wrapper2.call(-23.4, 35)).to.closeTo(-58.4, 0.001);
+            expect(wrapper2.call(-9.5, -2)).to.closeTo(-7.5, 0.001);
+        });
+
```
InverseIntegral, line 258: Add tests where you subtract a parameter from a double literal and vice-versa.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        it('should handle double multiplication', () => {
+            const wrapper = transpiler
+                .setSignature('mul', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function mul(a, b) { return a * b }');
+            wrapper.setFunctionName('mul');
+
+            expect(wrapper.call(3.45, 8.67)).to.closeTo(29.9115, 0.00001);
+            expect(wrapper.call(-10.48, 2.21)).to.closeTo(-23.1608, 0.00001);
```
InverseIntegral, line 288: Add a call where one parameter is zero.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+                .setSignature('mul', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32)
+                .transpile(content);
+            wrapper2.setFunctionName('mul');
+
+            expect(wrapper2.call(5.31, 4)).to.closeTo(21.24, 0.001);
+            expect(wrapper2.call(14.1, -3)).to.closeTo(-42.3, 0.001);
+            expect(wrapper2.call(-23.4, 35)).to.closeTo(-819, 0.001);
+            expect(wrapper2.call(-9.5, -2)).to.closeTo(19, 0.001);
+        });
+
```
InverseIntegral, line 334: Add tests where you multiply a parameter with a double literal.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+                .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32)
+                .transpile(content);
+            wrapper2.setFunctionName('div');
+
+            expect(wrapper2.call(5.31, 4)).to.closeTo(1.3275, 0.000001);
+            expect(wrapper2.call(14.1, -3)).to.closeTo(-4.7, 0.001);
+            expect(wrapper2.call(-23.4, 35)).to.closeTo(-0.6685714285714286, 0.000000000000001);
+            expect(wrapper2.call(-9.5, -2)).to.closeTo(4.75, 0.000000000000001);
+        });
+
```
InverseIntegral, line 409: Add tests where you divide a parameter by a double literal and vice-versa.

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+            expect(wrapper2.call([1, 2], 2.4)).to.closeTo(0.4166666666666667, 0.000000000000001);
+        });
+
+        it('should handle double division by 0', () => {
+            const wrapper = transpiler
+                .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function div(a, b) { return a / b }');
+            wrapper.setFunctionName('div');
+
+            expect(wrapper.call(2.52, 0.0)).to.equal(Infinity);
```
InverseIntegral, line 438: Integer division by zero throws an exception. Why is this not the case here?

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+        it('should handle double equality', () => {
+            const wrapper = transpiler
+                .setSignature('eq', WebAssemblyType.BOOLEAN, WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function eq(a, b) { return a == b }');
+            wrapper.setFunctionName('eq');
+
+            expect(wrapper.call(3.32, 3.32)).to.equal(true);
+            expect(wrapper.call(3.22, 2.54)).to.equal(false);
+        });
+
```
InverseIntegral, line 457: Add a test case where you check if `0.1 + 0.1 + 0.1 == 0.3` is `false`

test/transpiler/transpiler_double_test.ts
```diff
@@ -16,31 +16,814 @@ describe('Transpiler', () => {
...
+            expect(wrapper2.call(-23.4)).to.closeTo(-7.8, 0.001);
+        });
+
+        it('should handle double shorthand division with array', () => {
+            const wrapper = transpiler
+                .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY)
+                .transpile('function div(a) { var x = 1.13; x /= a[0]; return x; }');
+            wrapper.setFunctionName('div');
+
+            expect(wrapper.call([2, 3])).to.closeTo(0.565, 0.0001);
```
InverseIntegral: I couldn't find any test cases with for loops, while loops and function calls.

InverseIntegral (changes requested):
- Do while and for statements support the double type?
- The binaryen call expression uses an incorrect type
-  Is the property `variableTypes` of class `TypeInferenceVisitor` really necessary? You could use the `expressionTypes` property to get the same information.

test/transpiler/transpiler_double_test.ts
SeboCode: For consistency's sake and because imo you should never compare a double value without a delta, even if it is a non calculated value. But if you want it changed I wouldn't have a problem with that either.

test/transpiler/transpiler_double_test.ts
SeboCode: I just checked it with nodejs. This is the default behavior of JavaScript.

test/transpiler/transpiler_double_test.ts
SeboCode: I'll add tests for function calls. But I don't realy understand what you mean with for and while loop tests. If you mean something like this `while (x < 4.29) { x++; }`, the logic of the double value is already handled in the less then functionality. The while loop itself doesn't care if there is a double value used as it's parameter or not. It just expects a boolean value. Same goes for the for loop.

src/generator/type_inference_visitor.ts
SeboCode: No there is not. I just checked our code and it is robust enough, that this error never gets thrown, unless we expand the functionality. Therefore I think it is untestable.

SeboCode:
We could probably remove the `variableTypes` property. This would mean that we have to loop over all expressions, filter out the relevant ones and add them to a map. I think the code would look worse and be less performant than the solution we have now, which imo is quite good even though there is a little bit of redundancy in it.

test/transpiler/transpiler_double_test.ts
SeboCode: For consistency's sake and because imo you should never compare a double value without a delta, even if it is a constant value. But if you want me to change it, I wouldn't have a problem with that either.

SeboCode: For consistency's sake and because imo you should never compare a double value without a delta, even if it is a constant value. But if you want it changed I wouldn't have a problem with that either.

src/generator/generator_visitor.ts
```diff
@@ -63,8 +65,9 @@ class GeneratorVisitor extends Visitor {
     }
 
     protected visitNumericLiteral(node: NumericLiteral) {
-        const type = this.getExpressionType(node);
-        this.expressions.push(this.getOperationsInstance(type).const(node.value));
+        const instance = this.getOperationsInstance(this.getExpressionType(node));
+
```
InverseIntegral: Remove this empty line

src/generator/generator_visitor.ts
```diff
@@ -88,9 +91,9 @@ class GeneratorVisitor extends Visitor {
                 this.expressions.push(operand);
                 break;
             case '-':
-                const type = this.getExpressionType(node);
-                const operationsInstance = this.getOperationsInstance(type);
-                this.expressions.push(operationsInstance.sub(operationsInstance.const(0), operand));
+                const instance = this.getOperationsInstance(this.getExpressionType(node));
+
```
InverseIntegral: Remove this empty line.

src/generator/type_inference_visitor.ts
```diff
@@ -148,30 +149,38 @@ class TypeInferenceVisitor extends Visitor {
...
             this.expressionTypes.set(node.id, rightSideType);
-            this.updateVariableType(node.id, rightSideType);
+            this.updateIdentifierType(node.id, rightSideType);
         }
     }
 
     protected visitAssignmentExpression(node: AssignmentExpression): void {
         super.visit(node.right);
 
         if (isIdentifier(node.left)) {
```
InverseIntegral, line 159: Introduce a variable for `node.left`.

src/generator/type_inference_visitor.ts
```diff
@@ -148,30 +149,38 @@ class TypeInferenceVisitor extends Visitor {
...
+        }
+    }
 
-            if (node.operator !== '=') {
-                rightSideType = getCommonNumberType(this.getTypeOfIdentifier(node.left), rightSideType);
-                this.checkIfTypeIsUnchanged(node.left, rightSideType);
-            } else {
-                this.updateVariableType(node.left, rightSideType);
-            }
+    private assignToArray(memberExpression: MemberExpression, expression: Expression) {
```
InverseIntegral: Pass the node as one paramter and use a destructuring assignment in the method.

src/generator/type_inference_visitor.ts
```diff
@@ -148,30 +149,38 @@ class TypeInferenceVisitor extends Visitor {
...
+        super.visit(memberExpression);
 
-            this.expressionTypes.set(node.left, rightSideType);
-        } else if (isMemberExpression(node.left)) {
-            super.visit(node.left);
+        const rightSideType = this.getTypeOfExpression(expression);
+        this.expressionTypes.set(memberExpression, rightSideType);
+    }
+
+    private assignToIdentifier(identifier: Identifier, expression: Expression, operator: string) {
```
InverseIntegral: Pass the node as one paramter and use a destructuring assignment in the method.

src/generator/generator_visitor.ts
```diff
@@ -103,37 +106,27 @@ class GeneratorVisitor extends Visitor {
     protected visitBinaryExpression(node: BinaryExpression) {
         super.visitBinaryExpression(node);
 
-        let right = this.popExpression();
```
InverseIntegral: This new generic way of handling things introduces new special cases. Maybe there's a better solution to this. We can discuss it in person if you want to.

test/transpiler/transpiler_double_test.ts
```diff
@@ -149,13 +186,15 @@ describe('Transpiler', () => {
...
+            expect(wrapper.call(6.72, [3, 4])).to.closeTo(9.72, 0.001);
 
             const wrapper2 = transpiler
                 .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64)
                 .transpile('function add(a, b) { return a[0] + b; }');
             wrapper2.setFunctionName('add');
 
-            expect(wrapper2.call([1, 2], 2.4)).to.closeTo(3.4, 0.001);
+            expect(wrapper2.call([1, 2], 2.45)).to.closeTo(3.45, 0.001);
+            expect(wrapper2.call([3, 4], 9.22)).to.closeTo(12.22, 0.001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -119,27 +138,45 @@ describe('Transpiler', () => {
...
         });
 
+        it('should handle double addition parameter and literal', () => {
+            const wrapper = transpiler
+                .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function add(a) { return a + 3.48; }');
+            wrapper.setFunctionName('add');
+
+            expect(wrapper.call(6.81)).to.closeTo(10.29, 0.001);
+            expect(wrapper.call(11.76)).to.closeTo(15.24, 0.001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -119,27 +138,45 @@ describe('Transpiler', () => {
...
+
+            const wrapper2 = new Transpiler()
+                .setSignature('add', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function add(a) { return 5.79 + a; }');
+            wrapper2.setFunctionName('add');
+
+            expect(wrapper2.call(9.22)).to.closeTo(15.01, 0.001);
+            expect(wrapper2.call(4.63)).to.closeTo(10.42, 0.001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -168,27 +207,45 @@ describe('Transpiler', () => {
...
         });
 
+        it('should handle double subtraction parameter and literal', () => {
+            const wrapper = transpiler
+                .setSignature('sub', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function sub(a) { return a - 3.48; }');
+            wrapper.setFunctionName('sub');
+
+            expect(wrapper.call(6.81)).to.closeTo(3.33, 0.001);
+            expect(wrapper.call(11.76)).to.closeTo(8.28, 0.001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -168,27 +207,45 @@ describe('Transpiler', () => {
...
+
+            const wrapper2 = new Transpiler()
+                .setSignature('sub', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function sub(a) { return 5.79 - a; }');
+            wrapper2.setFunctionName('sub');
+
+            expect(wrapper2.call(9.22)).to.closeTo(-3.43, 0.001);
+            expect(wrapper2.call(4.63)).to.closeTo(1.16, 0.001);
```
InverseIntegral, line 232: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -215,6 +274,25 @@ describe('Transpiler', () => {
...
+        });
+
+        it('should handle double multiplication parameter and literal', () => {
+            const wrapper = transpiler
+                .setSignature('mul', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function mul(a) { return a * 3.48; }');
+            wrapper.setFunctionName('mul');
+
+            expect(wrapper.call(6.81)).to.closeTo(23.6988, 0.00001);
+            expect(wrapper.call(11.76)).to.closeTo(40.9248, 0.0001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -215,6 +274,25 @@ describe('Transpiler', () => {
...
+
+            const wrapper2 = new Transpiler()
+                .setSignature('mul', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function mul(a) { return 5.79 * a; }');
+            wrapper2.setFunctionName('mul');
+
+            expect(wrapper2.call(9.22)).to.closeTo(53.3838, 0.00001);
+            expect(wrapper2.call(4.63)).to.closeTo(26.8077, 0.00001);
```
InverseIntegral: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -247,13 +325,15 @@ describe('Transpiler', () => {
...
+            expect(wrapper.call(6.72, [3, 4])).to.closeTo(20.16, 0.001);
 
             const wrapper2 = transpiler
                 .setSignature('mul', WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64)
                 .transpile('function mul(a, b) { return a[0] * b; }');
             wrapper2.setFunctionName('mul');
 
-            expect(wrapper2.call([1, 2], 2.4)).to.closeTo(2.4, 0.001);
+            expect(wrapper2.call([1, 2], 2.45)).to.closeTo(2.45, 0.001);
+            expect(wrapper2.call([3, 4], 9.22)).to.closeTo(27.66, 0.001);
```
InverseIntegral, line 352: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -266,6 +346,24 @@ describe('Transpiler', () => {
...
         });
 
+        it('should handle double division parameter and literal', () => {
+            const wrapper = transpiler
+                .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function div(a) { return a / 3.48; }');
+            wrapper.setFunctionName('div');
+
+            expect(wrapper.call(6.81)).to.closeTo(1.956896551724138, 0.000000000000001);
+            expect(wrapper.call(11.76)).to.closeTo(3.379310344827586, 0.000000000000001);
```
InverseIntegral, line 374: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -266,6 +346,24 @@ describe('Transpiler', () => {
...
+
+            const wrapper2 = new Transpiler()
+                .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64)
+                .transpile('function div(a) { return 5.79 / a; }');
+            wrapper2.setFunctionName('div');
+
+            expect(wrapper2.call(9.22)).to.closeTo(0.6279826464208243, 0.000000000000001);
+            expect(wrapper2.call(4.63)).to.closeTo(1.250539956803456, 0.000000000000001);
```
InverseIntegral, line 383: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -296,13 +394,15 @@ describe('Transpiler', () => {
...
+            expect(wrapper.call(6.72, [3, 4])).to.closeTo(2.24, 0.001);
 
             const wrapper2 = transpiler
                 .setSignature('div', WebAssemblyType.FLOAT_64, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64)
                 .transpile('function div(a, b) { return a[0] / b; }');
             wrapper2.setFunctionName('div');
 
-            expect(wrapper2.call([1, 2], 2.4)).to.closeTo(0.4166666666666667, 0.000000000000001);
+            expect(wrapper2.call([1, 2], 2.45)).to.closeTo(0.4081632653061224, 0.000000000000001);
+            expect(wrapper2.call([3, 4], 9.22)).to.closeTo(0.3253796095444685, 0.000000000000001);
```
InverseIntegral, line 427: Add calls with negative numbers as well.

test/transpiler/transpiler_double_test.ts
```diff
@@ -825,5 +934,154 @@ describe('Transpiler', () => {
...
+                .transpile(content);
+            wrapper.setFunctionName('branch');
+
+            expect(wrapper.call(7.42)).to.equal(0);
+            expect(wrapper.call(4.15)).to.equal(1);
+            expect(wrapper.call(9.77)).to.equal(2);
+        });
+
+        it('should handle double in while loop', () => {
+            const content = 'function loop(a) { var times = 0, i = 0; while (i < a) { i++; times++; } return times; }';
```
InverseIntegral, line 991: Maybe add a test case where the counter of the while condition is a double.

InverseIntegral (changes requested):

test/transpiler/transpiler_double_test.ts
InverseIntegral: Remove it from the tests where it's not necessary.

src/generator/type_inference_visitor.ts
InverseIntegral: Wouldn't it be possible to test it with an unsupported statement / expression?

SeboCode: That would fail, because the `visit` method already throws an error if it should handle an unsupported statement / expression.

InverseIntegral (approved):

PR #54, opened 2018-12-01
Handle double arrays

InverseIntegral:


src/call_wrapper.ts
```diff
@@ -5,24 +5,60 @@ import Module = WebAssembly.Module;
...
+    private static fillMemory(memoryLayout: Map<number, number>, memory: WebAssembly.Memory) {
+        const i32Memory = new Uint32Array(memory.buffer);
+        const f64Memory = new Float64Array(memory.buffer);
+
+        for (const entry of memoryLayout.entries()) {
+            const key = entry[0];
+            const value = entry[1];
+
+            if (isOfType(value, WebAssemblyType.INT_32)) {
+                i32Memory[key  / 4] = value;
```
SeboCode: Add a constant for the offset and use it in the entire file.

src/call_wrapper.ts
```diff
@@ -5,24 +5,60 @@ import Module = WebAssembly.Module;
...
+            } else if (isOfType(value, WebAssemblyType.FLOAT_64)) {
+                f64Memory[key / 8] = value;
```
SeboCode: Add a constant for the offset and use it in the entire file.

src/call_wrapper.ts
```diff
@@ -5,24 +5,60 @@ import Module = WebAssembly.Module;
...
+        const memory = new Map<number, number>();
+
+        let wasmMemoryIndex = 0;
+
+        for (let i = 0; i < parameters.length; i++) {
+            const current = parameters[i];
+            const type = signature.parameterTypes[i];
+
+            if (type === WebAssemblyType.INT_32_ARRAY) {
+                memory.set(wasmMemoryIndex, current.length);
```
SeboCode: The code in both branches are the same. You could combine them by moving it out of the branches and add a new variable, that gets set to 4 resp. 8 in the branches.
We should also throw an error in the else part, if a new array type gets introduced.

Something like that:
```javascript
let offset;
if (Int) {
    offset = 4
} else if (Double) {
    if (!aligned) {...}
    offset = 8;
} else {
    throw new Error();
}
// combined logic
```

src/call_wrapper.ts
```diff
@@ -5,24 +5,60 @@ import Module = WebAssembly.Module;
...
+                    memory.set(wasmMemoryIndex, element);
+                    wasmMemoryIndex += 4;
+                }
+            } else if (type === WebAssemblyType.FLOAT_64_ARRAY) {
+                if (wasmMemoryIndex % 8 !== 0) {
+                    wasmMemoryIndex += (8 - (wasmMemoryIndex % 8));
+                }
+
+                memory.set(wasmMemoryIndex, current.length);
+                wasmMemoryIndex += 8;
```
SeboCode: Is there a reason why the double array has 8 bytes memory for it's length?

src/call_wrapper.ts
```diff
@@ -76,12 +112,25 @@ class CallWrapper {
...
+            let maxKey = [...memoryLayout.keys()].reduce((a, b) => Math.max(a, b));
+
+            if (isOfType(memoryLayout.get(maxKey), WebAssemblyType.INT_32)) {
+                maxKey += 4;
+            } else {
+                maxKey += 8;
+            }
+
+            const memory = new WebAssembly.Memory({
+                initial: Math.ceil(maxKey / Math.pow(2, 16)),
```
SeboCode: Extract the initial calculation into a separate method.

src/call_wrapper.ts
```diff
@@ -76,12 +112,25 @@ class CallWrapper {
...
+
+            const memoryLayout = tuple[0];
+            fixedParameters = tuple[1];
 
-            fixedParameters = CallWrapper.fillMemory(fixedParameters, writeableMemory);
+            let maxKey = [...memoryLayout.keys()].reduce((a, b) => Math.max(a, b));
+
+            if (isOfType(memoryLayout.get(maxKey), WebAssemblyType.INT_32)) {
+                maxKey += 4;
+            } else {
```
SeboCode: Use an else if here and maybe add an else part where an error gets thrown. This would be helpful when new types get introduced.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
     private getArrayElement(node: MemberExpression) {
         super.visitMemberExpression(node);
 
-        // The offset can not be used, because the memberaccess value can also be a mathematical term or a variable
-        return this.module.i32.load(0, 4, this.getPointer());
+        const type = this.getExpressionType(node);
+
+        switch (type) {
+            case WebAssemblyType.INT_32:
+                return this.module.i32.load(0, 4, this.getPointer(4));
```
SeboCode: Add a constant for the offset and use it in the entire file.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
+            case WebAssemblyType.FLOAT_64:
+                return this.module.f64.load(0, 8, this.getPointer(8));
```
SeboCode: Add a constant for the offset and use it in the entire file.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
+            default:
+                throw new Error(`Unknown type of member expression: ${type}`);
```
SeboCode: This only prints a number as the type in the error message. Use `WebAssemblyType[type]` instead.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
+
+        switch (type) {
+            case WebAssemblyType.INT_32_ARRAY:
+                const address = this.module.i32.sub(this.popExpression(), this.module.i32.const(4));
+                return this.module.i32.load(0, 4, address);
+            case WebAssemblyType.FLOAT_64_ARRAY:
+                const address2 = this.module.i32.sub(this.popExpression(), this.module.i32.const(8));
+                return this.module.i32.load(0, 4, address2);
+            default:
+                throw new Error(`Can\'t access length property of unknown type: ${type}`);
```
SeboCode: This only prints a number as the type in the error message. Use `WebAssemblyType[type]` instead.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
     private getArrayLength(node: MemberExpression) {
         this.visit(node.object);
 
-        const address = this.module.i32.sub(this.popExpression(), this.module.i32.const(4));
-        return this.module.i32.load(0, 4, address);
+        const type = this.getExpressionType(node.object);
+
+        switch (type) {
+            case WebAssemblyType.INT_32_ARRAY:
+                const address = this.module.i32.sub(this.popExpression(), this.module.i32.const(4));
```
SeboCode: Rename to `intArrayAddress`.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
+                return this.module.i32.load(0, 4, address);
+            case WebAssemblyType.FLOAT_64_ARRAY:
+                const address2 = this.module.i32.sub(this.popExpression(), this.module.i32.const(8));
```
SeboCode: Rename to `doubleArrayAddress`.

src/generator/generator_visitor.ts
```diff
@@ -427,23 +427,51 @@ class GeneratorVisitor extends Visitor {
...
+
+        switch (type) {
+            case WebAssemblyType.INT_32_ARRAY:
+                // @ts-ignore
+                return this.module.i32.store(0, 4, this.getPointer(4), value);
+            case WebAssemblyType.FLOAT_64_ARRAY:
+                // @ts-ignore
+                return this.module.f64.store(0, 8, this.getPointer(8), value);
+            default:
+                throw new Error(`Can\'t write to array with type: ${type}`);
```
SeboCode: This only prints a number as the type in the error message. Use `WebAssemblyType[type]` instead.

src/generator/type_inference_visitor.ts
```diff
@@ -132,7 +132,19 @@ class TypeInferenceVisitor extends Visitor {
...
+
+            switch (arrayType) {
+                case WebAssemblyType.INT_32_ARRAY:
+                    type = WebAssemblyType.INT_32;
+                    break;
+                case WebAssemblyType.FLOAT_64_ARRAY:
+                    type = WebAssemblyType.FLOAT_64;
+                    break;
+                default:
+                    throw new Error(`Unknown array type ${arrayType}`);
```
SeboCode: This only prints a number as the type in the error message. Use `WebAssemblyType[arrayType]` instead.

src/generator/type_inference_visitor.ts
```diff
@@ -159,18 +171,7 @@ class TypeInferenceVisitor extends Visitor {
...
-
-    private assignToArray(node: AssignmentExpression) {
-        const {left: memberExpression, right: expression} = node;
-
-        if (isMemberExpression(memberExpression)) {
-            super.visit(memberExpression);
-
-            const rightSideType = this.getTypeOfExpression(expression);
-            this.expressionTypes.set(memberExpression, rightSideType);
+            super.visit(node.left);
```
SeboCode, line 174: There is a check missing here. If you have an int array and make a shorthand assignment to a double value it should throw an error (use the `checkIfTypeIsUnchanged` functionality).
Please fix this error and add a test for this behavior.

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+        transpiler = new Transpiler();
+    });
+
+    describe('#transpile()', () => {
+
+        it('should handle array access', () => {
+            const wrapper = transpiler
+                .setSignature('array', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile('function array(arr) { return arr[0]; }');
+            expect(wrapper.setFunctionName('array').call([1.5, 2.0, 3.0])).to.equal(1.5);
```
SeboCode, line 20: Add more test cases (also one with negative numbers)

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+            expect(wrapper.call([11.1, 12.4, 13.7], 0)).to.equal(11.1);
+            expect(wrapper.call([14.2, 15.5, 16.8], 1)).to.equal(15.5);
+            expect(wrapper.call([17.3, 18.6, 19.9], 2)).to.equal(19.9);
+        });
+
+        it('should handle array access using expression', () => {
+            const wrapper = transpiler
+                .setSignature('array', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile('function array(arr) { return arr[1 + 2]; }');
+            expect(wrapper.setFunctionName('array').call([101.9, 102.8, 103.7, 104.6, 105.5])).to.equal(104.6);
```
SeboCode, line 41: Add more test cases

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+        });
+
+        it('should handle multiple arrays', () => {
+            const wrapper = transpiler
+                .setSignature('array', WebAssemblyType.FLOAT_64,
+                    WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile('function array(arr, arr2) { return arr2[2]; }');
+            expect(wrapper.setFunctionName('array').call([111.11, 112.12], [114.14, 115.15, 116.15])).to.equal(116.15);
```
SeboCode, line 52: Add more test cases

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+
+        it('should not modify array', () => {
+            const wrapper = transpiler
+                .setSignature('array', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile('function array(arr) { return arr[0]; }');
+
+            const array = [10.01, 20.01, 30.01];
+            wrapper.setFunctionName('array').call(array);
+
+            expect(array).to.eql([10.01, 20.01, 30.01]);
```
SeboCode, line 74: Add more test cases

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+        it('should handle mixed array parameters', () => {
+            const content = 'function arrayExport(arr1, arr2, arr3) { return arr1[0] + arr2[0] + arr3[0]; }';
+            const wrapper = transpiler
+                .setSignature('arrayExport', WebAssemblyType.FLOAT_64,
+                    WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile(content);
+
+            wrapper.setFunctionName('arrayExport');
+
+            expect(wrapper.call([5.5], [1, 2], [3.8, 4.4])).to.eq(10.3);
```
SeboCode, line 332: Add more test cases

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+        it('should handle mixed array parameters 2', () => {
+            const content = 'function arrayExport(arr1, arr2, arr3) { return arr1[0] + arr2[0] + arr3[0]; }';
+            const wrapper = transpiler
+                .setSignature('arrayExport', WebAssemblyType.FLOAT_64,
+                    WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.INT_32_ARRAY)
+                .transpile(content);
+
+            wrapper.setFunctionName('arrayExport');
+
+            expect(wrapper.call([1, 2], [5.5], [3, 4])).to.eq(9.5);
```
SeboCode: Add more test cases

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+                .setSignature('arrayExport', WebAssemblyType.FLOAT_64,
+                    WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.INT_32_ARRAY, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile(content);
+
+            wrapper.setFunctionName('arrayExport');
+
+            expect(wrapper.call([5.5], [1, 2], [3.8, 4.4])).to.eq(10.3);
+        });
+
+        it('should handle mixed array parameters 2', () => {
```
SeboCode: Wouldn't it be better to combine the two tests?

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -0,0 +1,352 @@
...
+
+            wrapper.setFunctionName('arrayExport');
+
+            expect(wrapper.call([], [5.5], [])).to.eq(1);
+            expect(wrapper.call([1], [5.5], [])).to.eq(2);
+            expect(wrapper.call([], [5.5], [2])).to.eq(2);
+            expect(wrapper.call([1], [5.5], [3])).to.eq(3);
+            expect(wrapper.call([1], [], [2])).to.eq(2);
+            expect(wrapper.call([1], [], [])).to.eq(1);
+            expect(wrapper.call([], [], [2])).to.eq(1);
```
SeboCode, line 368: Add one with all empty arrays

SeboCode (changes requested):
Is it possible to split the `call` method in the `CallWrapper` into multiple methods? Maybe into an import, execute and export method.

Add the new double array type to the api documentation.

src/call_wrapper.ts
InverseIntegral: Yes, it makes it easier for the alignment. If we used 4 bytes then there could be the case that we would have to round the memory address for the first element of the array. The access of the length would then have to subtract either -8 or -4 depending on the alignment which we can't know in the generator_visitor.

src/call_wrapper.ts
InverseIntegral: The else branch would never be reached because `isInteger` and `isDouble` already cover all possible numbers. The map `memoryLayout` only contains numbers (or undefined which throws an error in `isOfType`)

src/call_wrapper.ts
```diff
@@ -118,16 +124,16 @@ class CallWrapper {
             const memoryLayout = tuple[0];
             fixedParameters = tuple[1];
 
-            let maxKey = [...memoryLayout.keys()].reduce((a, b) => Math.max(a, b));
+            let maxAddress = [...memoryLayout.keys()].reduce((a, b) => Math.max(a, b));
```
SeboCode: Move the following logic into the `getAmountOfPages` method

src/generator/type_inference_visitor.ts
```diff
@@ -181,13 +190,13 @@ class TypeInferenceVisitor extends Visitor {
...
-                rightSideType = getCommonNumberType(this.getTypeOfIdentifier(identifier), rightSideType);
-                this.checkIfTypeIsUnchanged(identifier, rightSideType);
-            } else {
+            if (operator === '=') {
                 this.updateIdentifierType(identifier, rightSideType);
+            } else {
+                rightSideType = getCommonNumberType(this.getTypeOfIdentifier(identifier), rightSideType);
             }
 
+            this.checkIfTypeIsUnchanged(identifier, rightSideType);
```
SeboCode: This check is already performed in the if branch. Therefore move it into the else branch.

src/generator/type_inference_visitor.ts
```diff
@@ -220,11 +232,27 @@ class TypeInferenceVisitor extends Visitor {
...
+                    `from undefined to ${WebAssemblyType[type]}`);
+                } else {
+                    throw new Error(`Tried to change the type of ${name}` +
+                    `from ${WebAssemblyType[currentValue]} to ${WebAssemblyType[type]}`);
+                }
+            }
+        }
+    }
+
+    private checkIfTypeOfExpressionIsUnchanged(expression: Expression, type: WebAssemblyType) {
```
SeboCode: Remove this method

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -60,10 +69,17 @@ describe('Transpiler', () => {
                 .setSignature('array', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64_ARRAY)
                 .transpile('function array(arr) { return arr[0]; }');
```
SeboCode: This is somewhat of a useless test if the array doesn't get modified in it.

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -347,6 +366,27 @@ describe('Transpiler', () => {
...
+            const content = 'function arrayExport(arr) { ' +
+                'arr[0] = 3.14; return true;' +
+                '}';
+
+            expect(() => transpiler
+                .setSignature('arrayExport', WebAssemblyType.BOOLEAN, WebAssemblyType.INT_32_ARRAY)
+                .transpile(content)).to.throw();
+        });
+
+        it('should throw an error if the type of a double array is changed 2', () => {
```
SeboCode: Remove the 2 at the end of the name

SeboCode (changes requested):

test/transpiler/transpiler_double_array_test.ts
```diff
@@ -388,5 +388,18 @@ describe('Transpiler', () => {
...
+
+            const wrapper = transpiler
+                .setSignature('getSecond', WebAssemblyType.FLOAT_64, WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile(content);
+
+            wrapper.setFunctionName('getSecond');
+
+            expect(wrapper.call([1.5, 2.5])).to.eq(2.5);
+            expect(wrapper.call([1.5, 2])).to.eq(2);
+        });
```
SeboCode, line 403: Add a test where you try to add a double value to an int array.

SeboCode (changes requested):

SeboCode (approved):

PR #56, opened 2018-12-02
Add double benchmarks

SeboCode:
Added benchmarks:
- Sum double values
- Mergesort with double values
- Quicksort with double values
- Sum Array with double values


src/benchmark/cases/sum_array_double.js
```diff
@@ -0,0 +1,29 @@
...
+
+    for (var i = 0; i < array.length; i++) {
+        sum += array[i];
+    }
+
+    return sum;
+}
+
+function sumArrayFillDouble(array) {
+    for (var i = 0; i < array.length; i++) {
```
InverseIntegral, line 12: This  generates the sequence `[0, 0.5, 0.333, 0.25, 0.2]` . Wouldn't it be better if the numbers are slightly bigger? Maybe something like `[1, 1.5, 2, 2.5, 3, 3.5]` ?

src/benchmark/cases/quicksort_double.js
```diff
@@ -0,0 +1,104 @@
...
+    var current = 0.5;
+    var state = false;
+
+    while (i < array.length) {
+        array[i] = current;
+
+        current *= -1;
+
+        if (state) {
+            current++;
```
InverseIntegral: Maybe increase by `0.5` so that we also get integers in the array.

src/benchmark/benchmark.ts
```diff
@@ -27,8 +27,9 @@ class Benchmark {
         const times: Measurement[] = [];
 
         for (let i = 0; i < rounds; i++) {
+            const argsDeepCopy = JSON.parse(JSON.stringify(args));
```
InverseIntegral: You could use splice on all of the elements instead of using JSON.

src/benchmark/index.ts
```diff
@@ -82,66 +123,128 @@ const newtonsMethodFunc = {
...
+            returnType: WebAssemblyType.BOOLEAN }],
+        ['mergeSortFillInteger', { parameterTypes: [WebAssemblyType.INT_32_ARRAY],
+            returnType: WebAssemblyType.INT_32 }],
+        ['mergeSortWhileInteger', { parameterTypes: [WebAssemblyType.INT_32_ARRAY, WebAssemblyType.INT_32_ARRAY],
+            returnType: WebAssemblyType.BOOLEAN }],
+    ]),
+};
+
+const mergeSortDoubleFunc = {
+    arguments: [new Array(Math.pow(2, 20)).fill(0), new Array(Math.pow(2, 20)).fill(0)],
```
InverseIntegral: Don't fill the array explicitly here.

src/benchmark/index.ts
```diff
@@ -82,66 +123,128 @@ const newtonsMethodFunc = {
...
+            returnType: WebAssemblyType.BOOLEAN }],
+        ['mergeSortFillDouble', { parameterTypes: [WebAssemblyType.FLOAT_64_ARRAY],
+            returnType: WebAssemblyType.INT_32 }],
+        ['mergeSortWhileDouble', { parameterTypes: [WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.FLOAT_64_ARRAY],
+            returnType: WebAssemblyType.BOOLEAN }],
+    ]),
+};
+
+const quicksortIntegerFunc = {
+    arguments: [new Array(1000000)],
```
InverseIntegral: Is there a reason why this isn't 2^20 as well?

src/benchmark/cases/mergesort_double.js
```diff
@@ -0,0 +1,112 @@
...
+    var current = 0.5;
+    var state = false;
+
+    while (i < array.length) {
+        array[i] = current;
+
+        current *= -1;
+
+        if (state) {
+            current++;
```
InverseIntegral: See comment above.

test/time_limited/transpiler_test.ts
```diff
@@ -66,57 +83,115 @@ describe('Transpiler', function() {
...
+                    WebAssemblyType.FLOAT_64_ARRAY, WebAssemblyType.INT_32, WebAssemblyType.INT_32,
+                    WebAssemblyType.INT_32)
+                .setSignature('mergeSortIsSortedDouble', WebAssemblyType.BOOLEAN, WebAssemblyType.FLOAT_64_ARRAY)
+                .setSignature('mergeSortFillDouble', WebAssemblyType.INT_32, WebAssemblyType.FLOAT_64_ARRAY)
+                .setSignature('mergeSortWhileDouble', WebAssemblyType.BOOLEAN, WebAssemblyType.FLOAT_64_ARRAY,
+                    WebAssemblyType.FLOAT_64_ARRAY)
+                .transpile(content);
+            const parameters = [new Array(Math.pow(2, 20)), new Array(Math.pow(2, 20))];
+
+            expect(wrapper.setFunctionName('mergeSortWhileDouble').call(...parameters)).to.equal(true);
```
InverseIntegral: Why don't you pass the parameters directly?

InverseIntegral (changes requested):

src/benchmark/cases/sum_array_double.js
SeboCode: I think you misread the code here, because this generates the sequence `[0, 0.5, 1, 1.5, 2, 2.5...]`. It always uses the current index and the factor 0.5 in the multiplication.

